PR #29, opened 2015-05-18
translation update and extra translatable string

achumakov:
Thank you for prompt fix.
I've updated the translation, also revealing one more localizeable string.


sheabunge:
Ah, thanks for picking up the one I missed.

PR #35, opened 2015-05-20
An extra string translated

achumakov:
Seems now that's all Shea! Just extra 'Add new' translation for future versions. Thank you so much for such prompt response and version updates, it's really a pleasure!


sheabunge:
Thanks!

PR #45, opened 2016-05-14
Makes button "Dismiss" in feedback section translatable

pixolin:
Hi Shea,

this patch fixes a missing translation for the "Dismiss" button in the feedback section.

Best regards, Bego (@pixolin)


sheabunge:
Thank you!

PR #54, opened 2017-03-10
Visible Snippet Options - Please Review and Test SCSS Compilation.

CptCornWault:
This is just a simple improvement I use so that I can easily see which snippets are active and inactive.
I did this temporarily, and later I will make a sorting option.

If anyone else thinks it's useful, please implement it into the source.

### DIFF
--- a/css/menu-icon.scss
+++ b/css/menu-icon.scss
@@ -17,3 +17,37 @@
 	font-family: 'code-snippets' !important;
 	content: '\e600';
 }
+
+/* Added By CptCornWault - Visible Snippet Options SCSS
+.snippets .row-actions {
+	left: 0;
+	span {
+		opacity: .3;
+		left: 0;
+		&:first-of-type {
+			left: 0;
+			opacity: .5;
+		}
+	}
+}
+#the-list tr:hover span {
+	opacity: 1;
+}
+*/
+
+/* Added By CptCornWault - Visible Snippet Options CSS*/
+/* This is what the SCSS should compile to - compile untested - please remove this and uncomment the SCSS above after testing its compilation */
+.snippets .row-actions {
+	left: 0px;
+}
+.snippets .row-actions span {
+	opacity: .3;
+	left: 0px;
+}
+.snippets .row-actions span:first-of-type {
+	left: 0px;
+	opacity: .5;
+}
+#the-list tr:hover span {
+	opacity: 1;
+}


sheabunge:
I've checked the SCSS complation, and it does indeed compile as expected.

Very fancy! This is a nice-looking improvement. My only thought is: do you think that everybody would welcome it? I would only want to make this part of the plugin, and on by default, if the majority of people will want to use it. Otherwise, perhaps it could require changing an option to activate it.

CptCornWault:
Thanks for checking again. I was really stumped!

Sure, I'll add an option to turn it on and off.

Also, after writing this I discovered that that there is a default WP function that takes care of the issue I originally wrote it for.
I noticed that on this plugin, the snippets that are not active are missing the little blue horizontal bar to the left of the list item.

So, it seems like this addition is redundant in a way (even though I do like the links to always show on the list like this).
However, I could add some more edits that I like and some options to turn them on and off as well.
I want to take a stab at a few features for the code editor.

Do you want the option to toggle on/off to be in the settings menu as check boxes?
Shall I make them off by default?

I'll update when I can, cheers!




---- On Sun, 26 Mar 2017 03:46:43 -0700 Shea Bunge &lt;notifications@github.com&gt; wrote ----




I've checked the SCSS complation, and it does indeed compile as expected.

Very fancy! This is a nice-looking improvement. My only thought is: do you think that everybody would welcome it? I would only want to make this part of the plugin, and on by default, if the majority of people will want to use it. Otherwise, perhaps it could require changing an option to activate it.

sheabunge:
In the next version, I'm planning on [making a change on multisite](https://wordpress.org/support/topic/multisite-question-22/) to the snippets list to show network snippets on subsites, and as part of this it looks like I'll be making the row actions permanently visible.



So if that's what you're looking for, it's already in the works! I'll probably add a setting as well for those who hate the change and prefer them to only show on hover.

CptCornWault:
Ok, cool!
Feel free to use the code I sent if you want the fade/hover effect with it.

Cheers!

/j




---- On Tue, 04 Apr 2017 19:48:36 -0700 Shea Bunge &lt;notifications@github.com&gt; wrote ----




In the next version, I'm planning on making a change on multisite to the snippets list to show network snippets on subsites, and as part of this it looks like I'll be making the row actions permanently visible.



So if that's what you're looking for, it's already in the works! I'll probably add a setting as well for those who hate the change and prefer them to only show on hover.

sheabunge:
I definitely will! I might close this for now, but if I do end up implementing the fade and hover effects, I'll use your code and let you know. Thanks!

PR #78, opened 2019-01-19
Add search on line number feature

mjdewitt:
To help chase down PHP errors, I expanded the search capability in "search_callback()" to allow for narrowing search across snippets by line number in the code field only.  This is accomplished by parsing the search term for "@#:" and splitting the search term into three parts: <search term>@#:<line number>. Searches without "@#:" are handled normally.

NB: Since the search term is converted to entities when submitted, it creates a re-display problem of truncating the search term when the results are returned. I just convert $_REQUEST['s'] back to characters and didn't consider UTF-8. I'm not sure if this is a problem or not?

Hope you find this useful.

Mike

### DIFF
--- a/php/class-list-table.php
+++ b/php/class-list-table.php
@@ -986,22 +986,45 @@ class Code_Snippets_List_Table extends WP_List_Table {
 	 *
 	 * @return bool The result of the filter
 	 */
-	private function search_callback( $snippet ) {
-		static $term;
-		if ( is_null( $term ) ) {
-			$term = stripslashes( $_REQUEST['s'] );
+private function search_callback( $snippet ) {
+	static $term;
+	if ( is_null( $term ) ) {
+    		$_REQUEST['s'] = html_entity_decode( $_REQUEST['s'] );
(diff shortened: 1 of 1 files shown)


sheabunge:
Interesting idea! I like it, and I imagine it could be quite useful for tracking down specific errors.

I'm not sure that `@#:` is the best prefix, though. Aside from not being the easiest to remember the order, `#` being a special character in URLs could prove a little tricky.

Instead, how does something like `line:42` sound? Or is that too generic?

mjdewitt:
My concern with the search (delimiter?) was to pick something short that wouldn't likely occur in code/descriptions/titles. The idea behind @#: was to read it as "at line number" In the USA, "#" is used for the word, number. (Maybe this is universal?)

How about just "@:" or "\@:" ?

Not sure what is best, but glad you think this is interesting enough to consider.

sheabunge:
Ah right, that makes a lot of sense. I do quite like the clarity of actually saying `line` in the search keyword. Do you think that `@line:42` would work?

PR #92, opened 2020-02-26
"Only run once" msgid missing + IT transaltion

360fun:
Added "Only run once" msgid
Added translated msgstr

Warning: the original english string was missing, not only the translation, please check if other translations are miss it as well!


sheabunge:
Thank you for this!

PR #105, opened 2021-02-25
Extend search query for get_snippets

Spreeuw:
This adds two parameters to the `get_snippets()` args: `search` and `searchby`. This will make it easier (much faster) to retrieve snippets with a specific keyword. Here's an example of how this would be used:
```php
$snippets = get_snippets( [], null, [
	'search'      => 'do_shortcode',
	'searchby'    => ['code'],
	'active_only' => true,
	'orderby'     => 'priority',
	'limit'       => 5,
] );
```

I've taken the liberty to restructure the way the query is built up, using only a single call to `$wpdb->prepare()` at the end.
Ultimately this could be used to speed up the list-table search too, but because the code there is currently somewhat convoluted I didn't want to get into that before knowing whether this PR would be accepted (being able to use `get_snippets()` for search outside of the list-table actually serves my needs for now).

### DIFF
--- a/php/snippet-ops.php
+++ b/php/snippet-ops.php
@@ -37,6 +37,8 @@ function get_snippets( array $ids = array(), $multisite = null, array $args = ar
 		'limit'       => 0,
 		'orderby'     => '',
 		'order'       => 'desc',
+		'search'      => '',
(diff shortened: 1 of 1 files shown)


Spreeuw:
awesome, can't wait :)
Thanks for maintaining this super useful plugin 💘

PR #106, opened 2021-02-25
More action hooks

Spreeuw:
This PR makes it easier to extend/customize both the single edit and the list-table views by adding various action hooks and filters:

`apply_filters( 'code_snippets/admin/submit_actions', $actions, $snippet, $extra_actions )`
Allowing to add extra submit buttons to be added. I've also removed some code duplication from the  'extra save buttons' area by calling the `render_submit_buttons` method instead (allowing the same filter too hook in there).

`do_action( 'code_snippets/admin/before_title_input', $snippet );`
`do_action( 'code_snippets/admin/after_title_input', $snippet );`
`do_action( 'code_snippets/admin/manage/before_list_table' );`
Generic action hooks allowing custom elements to be inserted at the top parts of the pages too (a logical addition to the existing action hooks at the bottom)

```php
apply_filters( 'code_snippets/tag_it_options', array(
	'availableTags' => get_all_snippet_tags(),
	'allowSpaces'   => true,
) )
```
This is implemented using a `wp_localize_script` replacement for the current code using `wp_add_inline_script`, allowing for example to restrict the tags to a limited set, or expand them with a number of standard tags, as well as disabling spaces for tags (frontend only).

### DIFF
--- a/js/edit-tags.js
+++ b/js/edit-tags.js
@@ -9,8 +9,8 @@ import 'tag-it/js/tag-it';
 
 	try {
 		jQuery(tags_field).tagit({
-			availableTags: code_snippets_all_tags,
-			allowSpaces: true,
+			availableTags: code_snippets_tags.options.availableTags,
+			allowSpaces: code_snippets_tags.options.allowSpaces,
 			removeConfirmation: true,
 			showAutocompleteOnFocus: true,
 		});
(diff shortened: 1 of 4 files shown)


sheabunge:
Thank you so much for this! Excellent changes all around, I only had to make a few tweaks to bring them into line with some of the more recent changes. 

Please let me know if you would like to be credited on WordPress.org, and if so, what name, website, etc to use.

PR #118, opened 2021-08-26
Add Site Health integration

### LINKED ISSUE #117: Site Health Debug Information for Code Snippets
As a plugin developer/support member,
I want to know what code snippets customers use on their site,
So I can see their code snippet titles are and where they loads (admin/front),
And I can provide them with more detailed support by identifying snippets that might be causing a problem.

sc0ttkclark:
Fixes #117

Screencast of it in action: Screen Recording 2021-08-26 at 09 01 22 AM

This PR adds Code Snippet information to the Site Health Info tab to assist all plugin developers in their support efforts.

**Note:** Line breaks / HTML _are not supported on this screen_, so things are separated with a ` | \n`. This allows separation between the different active snippets. The line break doesn't show up on the screen but it _does show up on when clicking the copy button_ to provide to plugin/theme support teams. This technique has been helpful for other projects I have integrated with Site Health.

### DIFF
--- a/php/class-admin.php
+++ b/php/class-admin.php
@@ -17,6 +17,13 @@ class Admin {
 	public $menus = array();
 
 	/**
+	 * The Site Health class instance.
+	 * 
+	 * @var Site_Health
+	 */
+	public $site_health;
+
+	/**
 	 * Class constructor
 	 */
 	public function __construct() {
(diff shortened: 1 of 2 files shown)


sheabunge:
Thank you for this! It's a great idea that never occurred to me before.

I changed things up a bit so that the snippets are listed individually like plugins, but otherwise it's a fantastic addition.

Issue #98, opened 2020-07-13
Extend the UI slightly to add tabular forms ordered by areas of interest and sync


LebToki:
Hello there, 

What you guys have pulled so far is next to amazing.
What would be nice though is the ability to tab the snippets (move them according to their functions) to their focus areas.

You would end up with 
1. Wordpress UI ramp-ups
2. WooCommerce Specifics
3. Frontend CSS
4. Backend CSS
...and so on.

it could mean adding a few additional tabs/pages but on the long run will make our snippets more manageable especially when u are to deploy the next site

Would also be nice to fetch or sync all snippets from a central repository online (url of choice) 
once pegged to that file then any updates that target similar sites will use the central code base for those website such as electronics, spa, industrial and so on.

sheabunge:
Hi @LebToki,

Thanks for this suggestion! We're planning on adding something like this in v3.0 Pro, by dividing snippets into different types, and a cloud platform that can synchronize snippets is also in the works.

PR #124, opened 2022-02-24
Implement cloud storage for snippets

ancient-spirit:
> Thanks for this suggestion! We're planning on adding something like this in v3.0 Pro, by dividing snippets into different types, and a cloud platform that can synchronize snippets is also in the works.

See #98

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ts text eol=lf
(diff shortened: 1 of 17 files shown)


sheabunge:
Thank you for your contributions! Unfortunately, we're not looking to integrate a completely third-party service into the plugin at this time. I have merged the other contributions in this pull request, however. Hopefully the new hooks and methods will provide you with the basis to create third-party integrations in a separate project.

Issue #125, opened 2022-03-30
Enabling some common functionality and shortcuts


karmaral:
Hello!
I spend quite some time in your code editor, and I dearly miss some basic functionalities of my (or any, I'd argue) regular IDE.
These are: 
- `Ctrl + /` for comments
- `Alt + Up/Down` for swapping lines
- Code folding
- Vim keybindings

So far I managed to enable the first three no problem in a fork. I still have to dig in a little deeper regarding the custom keymaps.
Would these features be welcome if added?

Cheers

sheabunge:
Hi @karmaral,

That sounds super useful. Please feel free to create a PR with any changes to the existing keymap you think might be generally useful.

Also, I'm sure you're already across this, but the docs for the CodeMirror editor we are using has a section all about the different keymap options which should be useful: https://codemirror.net/doc/manual.html#keymaps

karmaral:
Alright, great!
Yeah, I was aware of that, it's just that the last time I wanted to add the keymaps manually into an npm setting I had issues with it. It seems they're already included in here so it is just a matter of exposing them. I'll open a PR soon with everything and we can take it from there.

sheabunge:
It's not a problem to add extra npm packages if you need them. The plugin has a build script that can grab the relevant parts and bundle it together for distribution on wordpress.org

karmaral:
Ok, so I was cleaning everything up, and suddenly it stopped working. I can't tell if it was after running the `package` command or what, but now Wordpress is not recognizing the plugin, even if I clone it from a clean fork. I'm really confused... have you ever had this issue?

I'll continue troubleshooting anyway.

karmaral:
Strange. It seems the lines `Author | Author URI` were duplicated in `code-snippets.php` and they were overriding `Plugin Name | Plugin URI`.
Reverting this fixes my issue, but I'm not sure how it managed to work on the first place before this. Anyways....

karmaral:
@sheabunge Let me know your thoughts on the PR when you can.

sheabunge:
Closing this as the PR is merged. Thank you again!

PR #126, opened 2022-04-03
Enable various common functionalities and shortcuts

karmaral:
This PR implements the functionalities discussed in #125:

- Toggle comments with `Ctrl + /`
- Swap lines up/down with `Alt + Up/Down`
- Enable code folding
- Enable keymaps included with the CodeMirror package (Vim, Emacs, Sublime)

Also fixes some duplicate plugin info attributes in `code-snippets.php` preventing the plugin to work.

I've made some comments on the changes too so we can discuss them in-place.

### DIFF
--- a/php/editor.php
+++ b/php/editor.php
@@ -31,6 +31,7 @@ function enqueue_code_editor( $type, $extra_atts = [] ) {
 
 	$default_atts = [
 		'mode'          => $modes[ $type ],
+		'inputStyle'    => 'textarea',
 		'matchBrackets' => true,
 		'extraKeys'     => [ 
 							'Alt-F' => 'findPersistent',
--- a/php/editor.php
+++ b/php/editor.php
@@ -35,6 +35,9 @@ function enqueue_code_editor( $type, $extra_atts = [] ) {
 		'extraKeys'     => [ 
 							'Alt-F' => 'findPersistent',
 							'Ctrl-Space' => 'autocomplete',
+							'Ctrl-/' => 'toggleComment',
+							'Alt-Up' => 'swapLineUp',
+							'Alt-Down' => 'swapLineDown',
 						],
 		'gutters'       => [ 'CodeMirror-lint-markers', 'CodeMirror-foldgutter'],
 		'lint'          => 'css' === $type || 'php' === $type,
--- a/php/editor.php
+++ b/php/editor.php
@@ -32,11 +32,15 @@ function enqueue_code_editor( $type, $extra_atts = [] ) {
 	$default_atts = [
 		'mode'          => $modes[ $type ],
 		'matchBrackets' => true,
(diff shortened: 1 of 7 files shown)


php/editor.php
```diff
@@ -32,11 +32,15 @@ function enqueue_code_editor( $type, $extra_atts = [] ) {
...
-		'gutters'       => [ 'CodeMirror-lint-markers' ],
+		'extraKeys'     => [ 
+							'Alt-F' => 'findPersistent',
+							'Ctrl-Space' => 'autocomplete',
+						],
+		'gutters'       => [ 'CodeMirror-lint-markers', 'CodeMirror-foldgutter'],
 		'lint'          => 'css' === $type || 'php' === $type,
 		'direction'     => 'ltr',
 		'colorpicker'   => [ 'mode' => 'edit' ],
+		'foldOptions' 	=> [ 'widget' => '...' ],
```
karmaral, line 47: The default widget looks really out of place. This looks much cleaner. The only caveat is that it must be manually added in the preview editor over at the settings page as well.

php/settings/settings-fields.php
```diff
@@ -215,6 +215,20 @@ function get_settings_fields() {
...
 			'codemirror' => 'styleActiveLine',
 		],
+		'keymap' => [
+			'name'       => __( 'Keymap', 'code-snippets' ),
+			'type'       => 'select',
+			'desc'       => __( 'The keymap to use in the editor.', 'code-snippets' ),
+			'default'    => 'default',
+			'options'    => [
+				'default' => __( 'Default', 'code-snippets' ),
+				'vim'     => __( 'Vim', 'code-snippets' ),
```
karmaral, line 225: Do you think these 3 are worth adding to the translation? Perhaps just having 'Default' is fine.

php/editor.php
```diff
@@ -31,6 +31,7 @@ function enqueue_code_editor( $type, $extra_atts = [] ) {
 
 	$default_atts = [
 		'mode'          => $modes[ $type ],
+		'inputStyle'    => 'textarea',
```
karmaral, line 34: I had to add this option (default seems to be `contentEditable`) because there appears to be some conflicts regarding the cursor and the vim keymap.

Also, there was a strange behavior (It doesn't seem to be present in the previous Code Snippets version) where selections would look like this
image

And using `'inputStyle' => 'textarea'` fixed it
image

The only caveat is that the default textarea seems to have a `rows=200` inline attribute that makes it super tall, and it will flash before it is initialized. It's probably something wordpress-related, but perhaps it can be fixed via css or some other way so it doesn't yank the layout.

I didn't do much digging on this, but I could if you think it's worth fixing.

karmaral:
Wonderful, thank you!

PR #163, opened 2023-02-16
The frontend method 'render_content_shortcode' has an 'eval' call tha…

JackSzwergold:
The frontend method `render_content_shortcode`has an `eval` call that adds two lines of whitespace that are not needed past `eval` processing so requesting the 'content' string be trimmed on return.

This caused issues with a website I am working on that  needs 100% empty values returned if no values exist in a code snippet. The source of the issue is these empty lines on line 206 which is `eval( "?>\n\n" . $snippet->code . "\n\n<?php" );` in the `render_content_shortcode` method contained in the `Frontend` class in the file `php/front-end/class-frontend.php`.

The solution I found is to just wrap the returned `$content` with `trim` like this: `return trim($content);`.

### DIFF
--- a/php/front-end/class-frontend.php
+++ b/php/front-end/class-frontend.php
@@ -225,7 +225,7 @@ class Frontend {
 			add_shortcode( self::CONTENT_SHORTCODE, [ $this, 'render_content_shortcode' ] );
 		}
 
-		return $content;
+		return trim($content);
 	}
 
 	/**


sheabunge:
This seems pretty reasonable to me. Alternatively, I suppose we could just not wrap the code in PHP tags if it's empty, I suppose. 

I added the newlines to ensure there's some whitespace in between the `<?php` and `?>` tags and the code. I'm not sure if it is technically necessary, but it felt best to be on the safe side.

JackSzwergold:
@sheabunge Yeah. I mean I am not that well versed with the history of the code to have even understood the value of the PHP wrapping. Thus, from my small slice of the perspective, a simple trim addresses the core issue I have faced and is quite harmless. Do as you wish based on what you know!

sheabunge:
Decided to leave the `trim`, but attach it to a filter hook so it can be removed or added to depending on the needs of each site. 

Thank you for your contribution!

JackSzwergold:
> Thank you for your contribution!

Thank you for the kind attention to solve this. Great work on your side!

PR #213, opened 2025-01-19
4 features to review in this PR

### LINKED ISSUE #148: Remove toggle hover effect
Hovering over a toggle makes it move. 
It would be better to not having this hover effect as it confuses a bit as it kinda starts to move on hover.
It should only move when one clicks the toggle.

### LINKED ISSUE #198: Allow nesting code snippet shortcodes within code snippet shortcodes
### Is your feature request related to a problem? Please describe

There is currently no way to use the `[code_snippet]` shortcode within a HTML snippet, even with  PHP and 'Evaluate additional shortcode tags' enabled.

### Describe the solution you'd like

Create a HTML snippet with the following code:

`[code_snippet id=42 php shortcodes]`

Create a second HTML snippet that references this first snippet:

`<div> [code_snippet id=1 php shortcodes] </div>`

The second snippet should render the content of the first by evaluating the shortcode.

### Describe alternatives you've considered

_No response_

### Additional context

Originally raised by [daisensou](https://wordpress.org/support/users/daisensou/) here: https://wordpress.org/support/topic/nesting-a-shortcode-inside-a-html-snippet/

### LINKED ISSUE #208: Disable scroll to top after save
Please add an option to disable scroll to top after saving a code snippet. Frequently I find myself need to scroll back down after every save to continue working in the section I was in, making tweaks on the fly. I know you have the add_action script to do this but would be nice if the scroll to top was disabled by default. Thanks guys, great work on Code Snippets, been loving it!!

lightbulbman:
@sheabunge @imantsk  - Please can you review the 4 changes made in this PR:

1. **Manage Snippets Page, effect of hover on toggle (linked to [issue 148](https://github.com/codesnippetspro/code-snippets/issues/148))** - user suggested that the movement of the toggle on hover is not a good UI experience and I agree, to enhance or give indication of interactivity I removed the movement on hover and added a background shadow. TBH most users will know what a toggle switch will do so we can even remove this altogether if feel there is no need.

2. **Nesting shortcodes (linked to [issue 198](https://github.com/codesnippetspro/code-snippets/issues/198) - further description there)** - by adding shortcodes [code-snippet "shortcodes"] without the "" in the code html snippet shortcode now allows the html snippet itself to contain another shortcode that will be rendered therefore allowing nested shortcodes. One limitation I just realised and haven't tested is that this may only allow one nested shortcode and not multiple, this could be an area of improvement for later on.

3. **Shortcode name attribute ([Issue 19](https://github.com/codesnippetspro/pro/issues/19))** - all html snippets where a shortcode is produced now include a name attribute which is the snippet name/ title capped to 3 words with trailing dots if longer to not make the shortcode so long...e.g. new format [code-snippets id=5 name='Test Snippet']

4. **Sticky admin notice (linked to [issue 208](https://github.com/codesnippetspro/code-snippets/issues/208) - principle based on user brandonjp suggestion)** - When saving a long snippet or any snippet the default behaviour is a scroll to top which shows the save admin notice, this change now removes this behaviour and instead makes the admin notice sticky so that it appears towards the top of the browser window so that if a user is editing a long snippet then they are not scrolled to top of page, and removes the notice automatically after 5 seconds.

### DIFF
--- a/src/css/edit.scss
+++ b/src/css/edit.scss
@@ -152,3 +152,9 @@ p.snippet-scope, .snippet-scope p {
 .wrap .notice {
 	scroll-margin: 0.75em;
 }
+
+#edit-snippet-form-container .cs-sticky-notice {
+	position: sticky;
+	top: 40px;
+	z-index: 100;
+}
--- a/src/css/manage.scss
+++ b/src/css/manage.scss
@@ -60,12 +60,6 @@ $inactive-color: #ccc;
 	height: 17px;
 	border-radius: 34px;
 	background-color: #ccc;
-	box-shadow: 0 0px 1px 0px rgba(0, 0, 0, .2);
-    transition: all 200ms ease-in-out !important;
-
-	&:hover {
-		box-shadow: 0 1px 3px 0px rgba(0, 0, 0, 0.5);
-	}
 
 	&::before {
 		transition: all .4s;
--- a/src/css/manage.scss
+++ b/src/css/manage.scss
@@ -60,6 +60,12 @@ $inactive-color: #ccc;
 	height: 17px;
 	border-radius: 34px;
 	background-color: #ccc;
+	box-shadow: 0 0px 1px 0px rgba(0, 0, 0, .2);
+    transition: all 200ms ease-in-out !important;
+
+	&:hover {
+		box-shadow: 0 1px 3px 0px rgba(0, 0, 0, 0.5);
+	}
 
 	&::before {
 		transition: all .4s;
@@ -72,19 +78,11 @@ $inactive-color: #ccc;
 		border-radius: 50%;
 	}
 
-	&:hover::before {
-		transform: translateX(40%);
-	}
-
 	.snippets .active-snippet & {
 		background-color: $active-color;
 
 		&::before {
 			transform: translateX(100%);
-		}
-
-		&:hover::before {
-			transform: translateX(60%);
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


imantsk (approved):
- https://github.com/codesnippetspro/code-snippets/issues/148 : I will approve for now but I'd like to hear what @JohnPixle thinks about the toggle style
 - https://github.com/codesnippetspro/code-snippets/issues/198 : tested, works well
 - https://github.com/codesnippetspro/pro/issues/19 : tested, works well
 - https://github.com/codesnippetspro/code-snippets/issues/208 : tested, nice improvement 👍

JohnPixle:
@imantsk Thanks for pinging me here.

One silly question, how can I test the file? 🙈  I tried to download the zip at the branch, but it did not work when I installed it.

From what I imagine, in worst case we could just remove the hover style of the toggle altogether, as Atif suggested. It's a very common UI element and I don't think it's critical to provide a hint for the fact that users can interact with it.

In any case, let me know how I can test it, happy to take a look!

sheabunge:
Some of these issues are eclipsed by the changes I've made when integrating the new UI, but I don't see anything wrong with making these changes for the current interface in the meantime.

I'm happy to merge this and include it in the new patch if we're happy with testing.

JohnPixle:
Thanks for helping me test this one.

Regarding the toggle, to be honest I am not a big fan of the shadow on hover. As @lightbulbman already mentioned:

> _TBH most users will know what a toggle switch will do so we can even remove this altogether if feel there is no need._

I agree with that. I believe our best bet is to simply remove the hover animation AND the background shadow. I don't think there is a need to imply interactivity here. So, to sum-up:

- No animation on hover
- No shadow on hover

Let me know your thoughts!
Thanks again

imantsk (approved):

PR #214, opened 2025-01-29
Patch for nested shortcode issues

lightbulbman:
As per issues raised on wordpress repo link below, I have amended the shortcode processing code. I did want to retain the functionality added as without it literal shortcodes i.e. [code_snippet id=x] placed in a HTML snippet won't do anything and I think its a valuable enough feature to add.

To summarise. the changes made were to retain the code I added and add the removed code back this way we can do both check for literal shortcodes aswell as handle php code in the HTML snippet where a shortcode may be generated within it

[https://wordpress.org/support/topic/3-6-7-broke-shortcode-evaluation/#post-18269725](https://wordpress.org/support/topic/3-6-7-broke-shortcode-evaluation/#post-18269725)

### DIFF
--- a/src/php/settings/settings-fields.php
+++ b/src/php/settings/settings-fields.php
@@ -8,6 +8,8 @@
 
 namespace Code_Snippets\Settings;
 
+use function Code_Snippets\code_snippets;
+
 /**
  * Retrieve the default setting values
  *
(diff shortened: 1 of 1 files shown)


imantsk:
📦 [code-snippets.3.6.8.zip](https://github.com/user-attachments/files/18585269/code-snippets.3.6.8.zip)

imantsk (approved):

src/php/front-end/class-front-end.php
```diff
@@ -329,28 +329,44 @@ public function render_content_shortcode( array $atts ): string {
...
 					}
 
+					// Increment depth and process the shortcode.
 					$depth++;
-					$atts = shortcode_parse_atts( $matches[1] );
-					$result = $this->render_content_shortcode( $atts );
-					$depth--;
+					$atts = shortcode_parse_atts($matches[1]);
 
+					// Ensure $atts is an array to avoid errors.
```
sheabunge: `shortcode_parse_atts` should always return an array – if we want to improve resiliency, we should be checking whether `$matches[1]` exists in the correct form before passing it.

sheabunge:
@imantsk what happened here? A force push? Weren't we going to revisit this?

imantsk:
@sheabunge this PR diverged from the core branch creating a merge conflict.
To not keep the PR pending i reconciled the branch with core using a force push.
we will surely revisit this but in a new branch and new PR.

PR #215, opened 2025-02-02
correcting laquo to lsaquo

brandonjp:
So minor, but this has bugged me for a couple years. This previous page button atop the table nav should be `&lsaquo;` instead of `&laquo;`
CleanShot 2025-02-02 at 14 27 00@2x-optimised png


imantsk (approved):

imantsk:
Beautifull! @brandonjp 👍 
Thank you a bunch for the pull request!

sheabunge:
Thank you for the contribution @brandonjp! I was a little confused when testing this at first, before I realised that this is occuring in the cloud snippets table only, not the local snippets table.

This fix should be in the next release.

PR #232, opened 2025-04-26
Merge similar translations strings

### LINKED ISSUE #231: Merge similar translations strings
The plugin has several similar translations strings, it make the translation harder. By merging these strings, we can reduce the total amount and simplify the translation process.

Image

Image

Image

Image

Image

Image

ramiy:
Fixes #231

### DIFF
--- a/src/php/admin-menus/class-manage-menu.php
+++ b/src/php/admin-menus/class-manage-menu.php
@@ -229,7 +229,7 @@ class Manage_Menu extends Admin_Menu {
 	 */
 	public function print_pro_message() {
 		if ( ! code_snippets()->licensing->is_licensed() ) {
-			echo '<span class="go-pro-badge">', esc_html_x( 'Pro', 'go pro badge', 'code-snippets' ), '</span>';
+			echo '<span class="go-pro-badge">', esc_html__( 'Pro', 'code-snippets' ), '</span>';
 		}
 	}
 
--- a/src/php/cloud/list-table-shared-ops.php
+++ b/src/php/cloud/list-table-shared-ops.php
@@ -115,11 +115,10 @@ function cloud_lts_build_action_links( Cloud_Snippet $cloud_snippet, string $sou
 				esc_html__( 'Download', 'code-snippets' )
 			) :
 			sprintf(
-				'<a class="cloud-snippet-download %s" href="%s" target="_blank"><span class="go-pro-badge">%s</span>%s</a>',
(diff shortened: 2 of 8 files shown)


src/php/admin-menus/class-manage-menu.php
```diff
@@ -229,7 +229,7 @@ protected function get_current_type(): string {
 	 */
 	public function print_pro_message() {
 		if ( ! code_snippets()->licensing->is_licensed() ) {
-			echo '<span class="go-pro-badge">', esc_html_x( 'Pro', 'go pro badge', 'code-snippets' ), '</span>';
```
sheabunge, line 232: I think we might want to keep this one as-is, as it is not a complete string, and `_x` provides context indicating that.

src/php/cloud/list-table-shared-ops.php
```diff
@@ -115,11 +115,10 @@ function cloud_lts_build_action_links( Cloud_Snippet $cloud_snippet, string $sou
 				esc_html__( 'Download', 'code-snippets' )
 			) :
 			sprintf(
-				'<a class="cloud-snippet-download %s" href="%s" target="_blank"><span class="go-pro-badge">%s</span>%s</a>',
+				'<a class="cloud-snippet-download %s" href="%s" target="_blank">%s</a>',
 				$additional_classes,
 				'https://codesnippets.pro/pricing/',
-				esc_html_x( 'Pro', 'pro only', 'code-snippets' ),
-				esc_html_x( ' Only', 'pro only', 'code-snippets' )
+				__( '<span class="go-pro-badge">Pro</span> Only', 'code-snippets' ),
```
sheabunge, line 121: Do we really want to include HTML in translation files?

ramiy: Yes, [we already do that in other strings](https://translate.wordpress.org/projects/wp-plugins/code-snippets/stable/en-au/default/?filters%5Bterm%5D=%3C&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&sort%5Bby%5D=original&sort%5Bhow%5D=asc&sorts=Apply+Sorting).

But the real reason is different. In current structure we hardcode the words order. In my language the words order should be reversed. The new string allows the translator to decide how to translate.

sheabunge: Makes sense! I still think it seems strange, but I accept that this is probably the better option.

PR #234, opened 2025-04-28
Use CSS logical properties to support both LTR & RTL

ramiy:
Add native browser LTR & RTL supports instead of using build-time scripts like `rtlcss`.

Replacing **physical properties** with **logical properties** and values adds flow relative direction instead of strict direction.

This PR prepares the ground for full RTL support with the browsers. When it happens, we will be able simply remove `rtlcss` and use the same CSS for both directions.

### DIFF
--- a/src/css/common/_editor.scss
+++ b/src/css/common/_editor.scss
@@ -26,13 +26,13 @@
 
 	&::before {
 		content: '';
-		padding-bottom: 5px;
+		padding-block-end: 5px;
 	}
 }
 
 /* Fix cursor color with rubyblue theme (see https://goo.gl/3HDgRm */
 .cm-s-rubyblue .CodeMirror-cursor {
-	border-left: 1px solid white !important;
+	border-inline-start: 1px solid white !important;
 }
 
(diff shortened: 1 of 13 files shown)


sheabunge:
Very interesting! Logical properties are a new concept to me, but they make a lot of sense.

Is browser support at a level where we would be fine to merge this now, or do we need a build step to improve compatability?

ramiy:
Good question. So the browser support is depends on the property you refer to. I used only [baseline](https://web.dev/baseline) **widely available** properties, meaning that they are not only fully interoperable by all browsers, but it's been over 30 months since the last browser started supporting the property. The PR doesn't include **newly available** not **limited available** properties.

As for the build step, we don't need to change anything at the moment. The build script creates a regular LTR file and a separate RTL file. When the PR will be merged, the too files will have less differences.

The goal is to make the LTR and RTL css files identical. Only then we will be able to remove `rtlcss` script and the CSS file will be bi-directional by default. We are not there yet. But PRs like this will help fulfill the goal.

PR #246, opened 2025-08-04
Feat/ Adjustable font size to code editor

CarolinaOP:
This pull request introduces the Adjustable Font Size feature for the Code Editor.

It adds a new option within the Editor Settings panel, allowing users to set a custom font size. The selected value is applied via inline CSS to the .CodeMirror class.

When the user updates the font size in the settings panel, the code preview dynamically reflects the change without requiring a page reload.

### DIFF
--- a/src/php/settings/settings-fields.php
+++ b/src/php/settings/settings-fields.php
@@ -162,12 +162,21 @@ function get_settings_fields(): array {
 			'codemirror' => 'indentUnit',
 			'min'        => 0,
 		],
+		'font_size'                   => [
+			'name'       => __( 'Font Size', 'code-snippets' ),
+			'type'       => 'number',
+			'label'      => _x( 'px', 'unit', 'code-snippets' ),
+			'codemirror' => 'fontSize',
+			'min'        => 8,
+			'max'        => 28,
+		],
 		'wrap_lines'                  => [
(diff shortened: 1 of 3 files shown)


src/php/settings/settings-fields.php
```diff
@@ -215,6 +216,14 @@ function get_settings_fields(): array {
 			'options'    => get_editor_theme_list(),
 			'codemirror' => 'theme',
 		],
+		'font_size'                   => [
+			'name'       => __( 'Code Editor Font Size', 'code-snippets' ),
```
imantsk: please rename this setting to just `Font Size` and move it right before `Wrap Lines`

Image

imantsk (changes requested):
just one small change request 🙏

src/php/settings/settings-fields.php
CarolinaOP: Ok, done. Rebuilt with that change. 

In my initial approach, I placed it there, but the user wouldn’t see the changes in the editor preview, since it’s not within the same viewport. That’s why I felt it made more sense to place it closer to the editor, allowing users to make fine adjustments to the font size and immediately see the effect.
Perhaps the three fields that allow fine-tuning (rather than just toggling a setting) would be better positioned near the editor for a more intuitive user experience.

PR #252, opened 2025-09-09
chore: automated tests

louiswol94:


test-playwright.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+if [ ! -d "node_modules" ]; then
+    echo "📦 Installing dependencies..."
+    npm install
+fi
+
+echo "🔨 Building plugin and installing PHP dependencies..."
+npm run build
+cd src && composer install && cd ..
```
sheabunge, line 18: `composer -d src install` is a little cleaner

test-playwright.sh
```diff
@@ -0,0 +1,43 @@
...
+
+echo "🚀 Starting WordPress environment..."
+npm run wp-env:start
+
+echo "⏳ Waiting for WordPress to start..."
+sleep 15
```
sheabunge, line 24: Is this going to be reliable? what are we waiting for here?

.wp-env.json
```diff
@@ -0,0 +1,17 @@
+{
+  "core": null,
+  "phpVersion": "8.1",
```
sheabunge, line 3: Can we test on a lower version? Ideally, we want to test on the minimum version we support.

sheabunge (approved):

Issue #275, opened 2025-10-26
Critical error with file-based execution


AmibeWebsites:
### Current behavior

File-based execution inserts the following security code at the top of each file:
`if ( ! defined( 'ABSPATH' ) ) { return; }`

This causes a fatal error with any script that uses namespace declarations.

> PHP Fatal error:  Namespace declaration statement has to be the very first statement or after any declare call in the script in [redacted]

### Expected behavior

Insert the security code after any namespace declaration statement.

### WordPress version

6.8.3

### Code Snippets version

3.8.0

### Code Snippets license

Core (free)

louiswol94:
Hi @AmibeWebsites 

Thanks for reporting this. I wonder what the best approach would be.

- For OOP snippets we would need to parse the top of the file, detect namespace/use statements and add this line after (if it does not exist already in the file). The solution becomes more complex.
- I feel like for OOP snippets the author should be responsible for adding this line, and we should only automatically guard procedural snippets.

Interested to hear your thoughts on this.

Kind regards
Louis

AmibeWebsites:
Hi Louis, thanks for your response.

In my snippets I already have the security line as a precaution - so in effect it would be duplicated, although I use a slightly different version.

The 'if it does not exist already' gets even more complex with variations in this code and coding style... but the duplication is not error generating, so I wouldn't be concerned about that.

Parsing and inserting code is common (hosts do it all the time with wp-config), typically with 'start' and 'end' comments to make it clear it was inserted, and for what purpose.

My approach would be to add a settings flag, defaulting to (parse and) 'insert' to cover yourselves, where authors then have the option to consent to turning it off and take on the responsibility.

(And to find a way to more gracefully handle the file version generating a critical error as per the plugins directory support item you're already aware of)

louiswol94:
Hi @AmibeWebsites , has the issue been resolved with the latest version?

PR #279, opened 2025-10-29
fix: improve file based execution with namespaced php snippets

louiswol94:


louiswol94:
Fixes https://github.com/codesnippetspro/code-snippets/issues/275

sheabunge:
Might be worth adding a specific filter (e.g. `code_snippets_php_snippet_file_header`) that just includes the code to be prepended, so it can be easily appended to if desired.

louiswol94:
> Might be worth adding a specific filter (e.g. `code_snippets_php_snippet_file_header`) that just includes the code to be prepended, so it can be easily appended to if desired.

Discussed offline ✅

sheabunge (approved):

PR #289, opened 2025-11-04
fix: share with subsites

imantsk:
- fixed styling and functionality for "Share with subsites" toggle in snippet editor
- fixed Shared snippets to appear in subsite snippets lists with individual activate/deactivate capability
- imporved visual indicators:
  - Network admin: Dashicons networking icon with "Shared with subsites" native tooltip
  - Subsites: Activation switches for shared snippets (instead of "Network Active" label)


src/php/class-list-table.php
```diff
@@ -307,7 +307,14 @@ protected function column_activate( Snippet $snippet ): string {
...
 
-		if ( $this->is_network && ( $snippet->shared_network || ( ! $this->is_network && $snippet->network && ! $snippet->shared_network ) ) ) {
+		// Show icon for shared network snippets on network admin.
+		if ( $snippet->shared_network && $this->is_network ) {
+			return '<span class="dashicons dashicons-networking network-shared" title="' . 
+				esc_attr__( 'Shared with Subsites', 'code-snippets' ) . 
+				'"></span>';
+		}
+
+		if ( $this->is_network && ( ! $this->is_network && $snippet->network && ! $snippet->shared_network ) ) {
```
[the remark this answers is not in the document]
louiswol94: @imantsk Please double check this condition ☝️

src/php/class-list-table.php
```diff
@@ -986,46 +989,43 @@ public function no_items() {
...
-
-			for ( $i = 0; $i < $limit; $i++ ) {
-				$snippet = &$snippets['all'][ $i ];
-
-				if ( in_array( $snippet->id, $ids, true ) ) {
+			// Mark shared network snippets on the network admin page.
+			foreach ( $all_snippets as $snippet ) {
+				if ( in_array( $snippet->id, $shared_ids, true ) ) {
 					$snippet->shared_network = true;
-					$snippet->tags = array_merge( $snippet->tags, array( 'shared on network' ) );
```
louiswol94, line 1013: Are you intentionally removing the tag?

src/php/rest-api/class-snippets-rest-controller.php
```diff
@@ -229,6 +230,36 @@ public function get_items( $request ): WP_REST_Response {
...
 
+	/**
+	 * Retrieve and merge shared network snippets.
+	 *
+	 * @param array<Snippet> $all_snippets List of snippets to merge with.
+	 * @param bool|null      $network      Whether fetching network snippets.
+	 *
+	 * @return array<Snippet> Modified list of snippets.
+	 */
+	private function get_network_items( array $all_snippets, $network ): array {
```
louiswol94, line 241: Maybe we can think about the method name. It does not just "get_network_items" :)

src/php/class-list-table.php
```diff
@@ -1062,7 +1062,7 @@ public function prepare_items() {
 		$snippets = array_fill_keys( $this->statuses, array() );
 
 		$all_snippets = apply_filters( 'code_snippets/list_table/get_snippets', get_snippets() );
-		$this->fetch_shared_network_snippets();
+		$all_snippets = $this->fetch_shared_network_snippets( $all_snippets );
```
louiswol94: Should the `apply_filters()` not be at the end?

louiswol94:
@imantsk It might be worth to try and get Playwright tests running on multisite as well. And then write a test for this case.

src/php/class-list-table.php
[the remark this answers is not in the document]
imantsk: fixed using the copilot suggestion

src/css/common/_badges.scss
```diff
@@ -18,7 +18,7 @@
 	padding-inline: 3px;
 	box-sizing: border-box;
 	gap: 5px;
-	line-height: 1;
+
```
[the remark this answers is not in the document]
imantsk: reverted.

src/php/rest-api/class-snippets-rest-controller.php
imantsk: do you have any good suggestion? `get_network_snippets` maybe?

src/php/class-list-table.php
imantsk: yes, cos i dont think its necessary to temper with user assigned tags to snippets. however i think i understand the reason it was here, which probably is to indicate on the subsite that this snippet is coming from the the network.
i see that currently these snippets have the same action links `Edit | Clone | Export | Trash` which lead back to the main network site and this might pe confusing.
to address this ill try to implement the following:
- a snippets list filter for network shared snippets
- on subsites, substitute the row actions to clearly show that this is a network shared snippet and render only the Edit action link, only if the current user `is_super_admin`

src/css/common/_badges.scss
```diff
@@ -27,6 +27,13 @@
 	}
 }
 
+.network-shared {
+  color: #2271b1; 
+  font-size: 22px;
+  width: 100%;
+  cursor: help;
```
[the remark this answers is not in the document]
imantsk: fixed

src/php/class-list-table.php
imantsk: fixed

src/php/class-list-table.php
imantsk: implemented both features

imantsk:
> @imantsk It might be worth to try and get Playwright tests running on multisite as well. And then write a test for this case.

Very good point @louiswol94! However, since this is quite a substantial task, we will address it in a separate issue: https://github.com/codesnippetspro/code-snippets/issues/292

louiswol94 (approved):

PR #298, opened 2025-11-13
Migrate to native CSS direction management

ramiy:
Delete `RtlCssPlugin`, we no longer need it. Bi-directional support can be achieved using native CSS logical properties.


src/css/common/_switch.scss
```diff
@@ -40,13 +40,13 @@ input[type='checkbox'].switch:checked {
 
 	&::before {
 		background-color: white;
-		transform: translateX(100%);
+		transform: translateX(calc(100% * var(--cs-direction-multiplier)));
```
sheabunge: Interesting!

sheabunge:
Definitely nice to remove some unnecessary code and packages!

PR #299, opened 2025-11-13
Fix the position of "code direction" field in RTL mode

ramiy:
In RTL, the "code direction" field should be inline.

Before:

Screenshot 2025-11-13 at 15 40 46

After:

Screenshot 2025-11-13 at 15 42 02


sheabunge:
Good catch!

JohnPixle:
sorry I'm late, but it's looking good 👍🏻

PR #302, opened 2025-11-14
Change tested up to version from 6.8.2 to 6.8

ramiy:
The version number should only include major versions 6.8.

image


imantsk:
@ramiy good catch! can you tell me with is the dashboard in your screenshot? 
fyi, since the change is small and safe is, for this time, ill update the pr base to `core`. 
Im about to do a new release and will take the chance to push this live directly :)

ramiy:
@imantsk the screenshot is from https://wordpress.org/plugins/plugin-check/

install the plugin and run all the tests. You will see many issues that need attention.

PR #304, opened 2025-11-16
fix: tinymce plugin issue

imantsk:
## TL;DR
- Improved vendor dependency management (PRO)  
- Enhanced UI for cloud snippets  
- Refined styling for back navigation  

## What changed?
- Removed `.browserslistrc` file to prevent compatibility issues  
- Improved TinyMCE integration with custom scissors icon  
- Enhanced Imposter plugin integration to namespace-prefix vendor dependencies under `Code_Snippets\Vendor\` to prevent conflicts with other plugins  
- Updated autoloader to dynamically remove original namespace PSR-4 mappings at runtime  
- Enhanced cloud snippets UI with improved card layout, better action buttons, and fixed pagination  
- Added back navigation button styling with arrow indicator  

## How to test?
- Test the cloud snippets interface to verify the improved card layout and pagination  
- Check the back navigation from the snippet edit screen  
- Verify the TinyMCE integration with the new scissors icon  
- Test with a plugin that uses the same libraries to ensure no conflicts occur (PRO)  
- Check RTL support for proper display of UI elements


src/css/manage/_cloud.scss
```diff
@@ -323,22 +362,22 @@ td.column-download {
 
 @keyframes thumb {
 	0% {
-		transform: rotate(0)
+		transform: rotate(0);
 	}
 
 	33% {
-		transform: rotate(calc(7deg * var(--cs-direction-multiplier)));
+		transform: rotate(7deg);
```
sheabunge, line 369: What's the purpose behind removing these changes? Were they causing issues?

src/js/mce.ts
```diff
@@ -93,16 +93,40 @@ export const insertSourceMenu = (editor: Editor, ed: LocalisedEditor) => ({
 	}
 })
 
-tinymce.PluginManager.add('code_snippets', function (editor) {
+// Custom scissors icon as base64-encoded SVG (same as used in WP admin menu)
+// Base64-encoded version of menu-icon.svg
+const scissorsIcon =
+	'data:image/svg+xml;base64,' +
+	'PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGZpbGw9Im5vbmUiIHZpZX' +
```
sheabunge, line 100: I'm a little concerned this dump of base64 code might cause false-positives in security plugins (my concerns may be unfounded)

sheabunge (approved):

src/css/manage/_cloud.scss
imantsk: No particolar reason. I just synced back the styles from what was in the pro repo. Also maybe a bit easier to understand

src/js/mce.ts
imantsk: I taught about that too.. i wanted to make a function that takes the actual file and encodes it in base64 ar required but somethig wasnt working right, so i ended up doing it like this.. maybe you could help with that?

PR #307, opened 2025-11-25
Add promotions

ramiy:
image

### DIFF
--- a/src/php/promotions/elementor-pro.php
+++ b/src/php/promotions/elementor-pro.php
@@ -1,0 +1,153 @@
+<?php
+namespace Code_Snippets\Promotions;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Elementor Pro promotion class.
+ */
+class Elementor_Pro {
+
+	/**
+	 * Class constructor.
+	 */
+	public function __construct() {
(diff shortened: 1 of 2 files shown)


src/php/promotions/elementor-pro.php
```diff
@@ -0,0 +1,130 @@
...
+		?>
+		<div class="notice notice-info is-dismissible">
+			<p>
+				<strong><?php esc_html_e( 'Looking for a better way to manage your custom code?', 'code-snippets' ); ?></strong>
+			</p>
+			<p>
+				<?php esc_html_e( 'Code Snippets Pro provides a powerful and user-friendly alternative to Elementor Custom Code, with cloud sync, advanced features, and an intuitive interface.', 'code-snippets' ); ?>
+			</p>
+			<p>
+				<a href="<?php echo esc_url( admin_url( 'admin.php?page=snippets' ) ); ?>" class="button button-primary">
```
sheabunge, line 39: Please use the proper API functions for generating plugin page URLs (e.g. `Plugin::get_menu_url()`)

src/php/promotions/elementor-pro.php
sheabunge, line 1: I'm not 100% sure why this is its own class and sub-namespace. Shouldn't it be under `admin/`? I also don't think it needs to be in the core plugin – it could be unique in Code Snippets Pro.

PR #309, opened 2025-11-27
internal: add phpunit tests

louiswol94:


tests/install-wp-tests.sh
```diff
@@ -0,0 +1,170 @@
...
+
+download() {
+    if [ `which curl` ]; then
+        curl -s "$1" > "$2";
+    elif [ `which wget` ]; then
+        wget -nv -O "$2" "$1"
+    fi
+}
+
+if [[ $WP_VERSION =~ ^[0-9]+\.[0-9]+\-(beta|RC)[0-9]+$ ]]; then
```
ramiy: Just a question: Is this really necessary? do you really check beta/RC versions?

louiswol94: @ramiy Good question. I got this from the [standard WP-CLI scaffold script](https://github.com/wp-cli/scaffold-command/blob/730dbe73fdbeab454ef3907d272b020575361e02/templates/install-wp-tests.sh#L39). But I simplified it now. Thanks!

.github/workflows/phpunit.yml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-22.04
+    name: PHPUnit - Test Results Summary
+    steps:
+      - name: Test status summary
+        run: |
+          echo "PHP 7.4: ${{ needs.phpunit-php-7-4.result }}"
+          echo "PHP 8.0: ${{ needs.phpunit-php-8-0.result }}"
+          echo "PHP 8.1: ${{ needs.phpunit-php-8-1.result }}"
+          echo "PHP 8.2: ${{ needs.phpunit-php-8-2.result }}"
+          echo "PHP 8.3: ${{ needs.phpunit-php-8-3.result }}"
```
ramiy, line 75: Maybe we should add 8.4?

louiswol94: @ramiy Yes sure! Updated.

.github/workflows/phpunit-test.yml
```diff
@@ -0,0 +1,90 @@
+name: PHPUnit Test Runner
+
+on:
+  workflow_call:
+    inputs:
+      php-version:
+        required: false
+        type: string
+        default: '8.1'
```
ramiy:
```suggestion
        default: '8.2'
```

The _most used_ PHP version in WordPress websites is PHP 8.2.

Moreover, by the end of November 2025, PHP 8.1 will reach its _End of Life_.

Ref:
* https://wordpress.org/about/stats/
* https://www.php.net/supported-versions.php

ramiy (changes requested):

ramiy (approved):

PR #319, opened 2025-12-11
chore: satisfy plugin checks for WP core 6.9

imantsk:
- [x] Add required translators comments for placeholder strings
- [x] Escape translatable output in AJAX/version switch/import flows
- [x] Escape cloud search list action links
- [x] Prefix globals/locals to satisfy naming checks in loader and views


src/php/admin-menus/class-import-menu.php
```diff
@@ -128,30 +128,36 @@ protected function print_messages() {
...
+				echo '<div id="message" class="updated fade"><p>';
+
+				$imported = intval( $_REQUEST['imported'] );
+
+				if ( 0 === $imported ) {
+					esc_html_e( 'No snippets were imported.', 'code-snippets' );
+
+				} else {
+					$imported_count = sprintf( '<strong>%s</strong>', esc_html( number_format_i18n( $imported ) ) );
+					printf(
```
sheabunge: `printf` seems redundant here. What are we intending to achieve with this change?

src/php/cloud/class-cloud-search-list-table.php
```diff
@@ -173,11 +173,11 @@ public function display_rows() {
...
 							echo '</a>';
 							?>
-						</h3>
-						<ul class="action-buttons">
-							<?php echo cloud_lts_build_action_links( $item, 'search' ); ?>
-						</ul>
-					</div>
+							</h3>
+							<ul class="action-buttons">
+								<?php echo wp_kses_post( cloud_lts_build_action_links( $item, 'search' ) ); ?>
```
sheabunge: I believe this will break the thickbox script. Escaping isn't necessary here – if we want to improve compliance, the underlying function really should be refactored to do output itself.

src/php/settings/class-version-switch.php
```diff
@@ -230,17 +230,18 @@ public static function handle_version_switch( string $target_version ): array {
 			return self::create_error_response( $install_result->get_error_message() );
 		}
 
-		if ( $install_result ) {
-			delete_transient( VERSION_CACHE_KEY );
+			if ( $install_result ) {
```
sheabunge: Looks like this introduces some incorrect indentation.

src/php/views/partials/cloud-search.php
```diff
@@ -14,8 +14,8 @@
  * @var Manage_Menu $this
  */
 
-$search_query = isset( $_REQUEST['cloud_search'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['cloud_search'] ) ) : '';
```
sheabunge: Why rename these?

src/php/views/partials/cloud-search.php
```diff
@@ -30,32 +30,32 @@
 <form method="get" action="" id="cloud-search-form">
 	<?php List_Table::required_form_fields( 'search_box' ); ?>
 	<label class="screen-reader-text" for="cloud_search">
-		<?php esc_html_e( 'Search cloud snippets', 'code-snippets' ); ?>
```
sheabunge: Introduces incorrect indentation.

sheabunge (changes requested):

PR #321, opened 2025-12-16
fix: multisite issues

### LINKED ISSUE #311: Network Settings Option to restrict Snippets Menu doesn't work
### Current behavior

In a WordPress Multisite network running the latest versions of WordPress and the plugin, it is currently not possible to hide the Snippets menu from site administrators.

Unchecking the “Snippets” option under Network Admin → Settings → Menu Settings should limit Snippets access to Network Administrators (Super Admins) only. However, this setting appears to have no effect—regular site administrators can still see and access the Snippets menu on their individual subsites.

### Expected behavior

Expected Behavior: Unchecking the “Snippets” box in Network Admin → Settings → Menu Settings should restrict access to all Snippets menus to Network Administrators only.

Documentation - https://codesnippets.pro/doc/how-does-code-snippets-work-on-wordpress-multisite/

### Steps to reproduce

_No response_

### WordPress version

6.9

### Code Snippets version

3.9.2

### Code Snippets license

Core (free)

### Anything else?

_No response_

imantsk:
Fixes #311 #317

### Summary
- Fixes Multisite *“Menu Settings → Snippets”* not restricting subsite access by enforcing the network capability when the subsite Snippets menu is disabled (`src/php/class-plugin.php`).
- Fixes trashed network snippets executing on Multisite by ensuring `active = -1 (trashed)` never qualifies as executable when building the active-snippet list (`src/php/class-db.php, src/php/flat-files/classes/class-snippet-files.php`).
- Fixes file-based execution *“Cannot redeclare …”* collisions in Network Admin by ensuring the “local site” flat-file index uses the site table (not the network table) so the same network snippet isn’t loaded twice (`src/php/flat-files/classes/class-snippet-files.php`).

### Testing

#317 (trashed network snippets executing / file-based collision)
1. On Multisite, create a Network snippet using the issue’s “green banner” code and Save and Activate → banner appears.
```php
add_action('wp_footer', 'snippet_active_banner');
add_action('admin_footer', 'snippet_active_banner');
function snippet_active_banner() {
    echo '<div style="position:fixed;bottom:0;left:0;right:0;background:#4caf50;color:#fff;text-align:center;padding:10px;font-weight:600;z-index:999999;">✓ This Code Snippet is active!</div>';
}
```
2. Deactivate → banner disappears.
3. Trash the snippet (inactive) → banner stays gone; confirm it does not execute while in Trash.
4. Restore → snippet returns as inactive; banner still gone until re-activated.
5. Repeat steps 1–3 but trash while active → banner stops immediately.
6. Enable file-based execution, repeat steps 1–5, and confirm no fatal Cannot redeclare ... occurs.

#311 (Menu Settings restriction)
1. Network Admin → Settings → Network Settings → Menu Settings: uncheck Snippets, save.
2. Log in as a regular subsite admin → Snippets menu is hidden.
3. Direct URL check as subsite admin: `wp-admin/admin.php?page=snippets` / `wp-admin/admin.php?page=add-snippet` should be blocked.
4. Re-check Snippets in Menu Settings → subsite admin can see/access Snippets again.


src/php/flat-files/classes/class-snippet-files.php
```diff
@@ -408,7 +595,17 @@ public function add_settings_fields( array $fields ): array {
...
+	/**
+	 * Recreate all flat files when file-based execution settings are updated.
+	 *
+	 * @param array<string, mixed> $settings Settings data.
+	 * @param array<string, mixed> $input    Raw input data.
+	 *
+	 * @return void
+	 */
 	public function create_all_flat_files( array $settings, array $input ): void {
+		unset( $input );
```
louiswol94: Why is this needed?

src/php/class-db.php
```diff
@@ -197,53 +197,95 @@ public static function create_table( string $table_name ): bool {
...
-		$cache_key = sprintf( 'active_snippets_%s_%s', sanitize_key( join( '_', $scopes ) ), $table_name );
-		$cached_snippets = wp_cache_get( $cache_key, CACHE_GROUP );
-
-		if ( is_array( $cached_snippets ) ) {
-			return $cached_snippets;
-		}
-
-		if ( ! self::table_exists( $table_name ) ) {
+	public static function is_network_snippet_enabled( int $active_value, int $snippet_id, array $active_shared_ids ): bool {
+		if ( -1 === $active_value ) {
```
louiswol94, line 280: Just double checking - how this became an int?

louiswol94 (approved):

src/php/flat-files/classes/class-snippet-files.php
imantsk: this change is coming from the automated PHPCS fixing.
i believe that the 2 arg method signature was implemented due to the `code_snippets/settings_updated` action hook called in `Snippet_Files::register_hooks()` but effectively `array $input` is never used .
for this reason, in order to satisfy the `UnusedFunctionParameter` issue, my autofixer added this `unset` call.
maybe its fine to simply remove the method arg and adjust the hook registration?

src/php/class-db.php
imantsk: i think i explicitly `intval()` the `$active_value` value before calling this method. 
the incoming value is stored as a string in the db, but since we don’t use `declare(strict_types=1)` in our plugin, with this typehint, i can rely on PHP to coerce it to an integer, preventing the `-1` truthiness issue (root cause of #317)

PR #328, opened 2026-01-17
[CWA-269] feat: Native Admin Bar QuickNav (Snippets) by DECKERWEB

imantsk:
## Summary
This PR brings a native **Snippets QuickNav** experience into Code Snippets via the WordPress Admin Bar. It adds a dedicated “Snippets” top-level admin bar menu with quick links plus paginated Active/Inactive snippet lists that update dynamically.

## Credit
This work is based on the approach and UX established by **“Snippets QuickNav”** by **David Decker – DECKERWEB** (GitHub: @deckerweb):
https://github.com/deckerweb/snippets-quicknav
This PR implements the feature natively, following Code Snippets’ existing architecture and coding style.

## What’s included
- **Admin Bar → Snippets** menu using the existing scissors icon (styled to better match WP admin bar icons)
- Main nodes:
  - Manage
  - Add New
  - Import
  - Settings
  - Active Snippets
  - Inactive Snippets
- Active/Inactive lists:
  - Show snippet titles prefixed with type, e.g. `(PHP) My snippet title` (type inferred from `type`)
  - Paginate when the list exceeds the configured limit
  - Pagination controls live at the top of the submenu and fetch pages via the Snippets REST API
- **Safe Mode**
  - Safe Mode indicator shows only when Safe Mode is active
  - Adds a Safe Mode docs link as a submenu item: `https://snipco.de/safe-mode` (opens in new tab, uses Dashicons)
- **Free vs Pro gating**
  - Pro-only “Add New” types (CSS/JS/Conditions) are greyed out in free and link to the upgrade page

## Settings added
- Toggle to enable/disable the admin bar menu (default: enabled)
- “Snippets per page in Admin Bar” (default: 20)

## How to test
1. Go to Code Snippets settings and confirm Admin Bar QuickNav is enabled; set the per-page limit to something small (e.g. 5).
2. Open the WP Admin Bar → **Snippets** and verify the main nodes and submenu structure.
3. Open **Active Snippets** / **Inactive Snippets** and click pagination controls; confirm the list updates correctly.
4. Enable Safe Mode and confirm the Safe Mode indicator appears and the Safe Mode docs submenu link opens in a new tab.


src/js/admin-bar.ts
```diff
@@ -0,0 +1,296 @@
+type PaginationStatus = 'active' | 'inactive'
```
sheabunge: Please make sure that types which are referenced in exported functions are also exported.

src/js/admin-bar.ts
```diff
@@ -0,0 +1,296 @@
...
+type PaginationAction = 'first' | 'prev' | 'next' | 'last'
+
+type SnippetResponseItem = {
```
sheabunge: Object types like these should ideally be `interface`s for consistency.

src/js/entries/admin-bar.ts
```diff
@@ -0,0 +1,296 @@
...
+
+const setLoading = (controls: HTMLElement, loading: boolean) => {
+	controls.dataset.loading = loading ? 'true' : 'false'
+}
+
+const buildRequestUrl = (status: PaginationStatus, page: number): string => {
+	if (!config) return ''
+
+	const url = new URL(config.restUrl)
+	url.searchParams.set('status', status)
```
sheabunge, line 67: nice use of `URL`.

src/js/admin-bar.ts
```diff
@@ -0,0 +1,296 @@
...
+}
+
+const buildSnippetPlaceholder = (snippetId: number): string => {
+	if (!config?.snippetPlaceholder) return `Snippet #${snippetId}`
+
+	return config.snippetPlaceholder.replace(/%(\d+\$)?d/, String(snippetId))
+}
+
+const getTypeFromScope = (scope: string): string => {
+	if (scope.endsWith('-css')) return 'css'
```
sheabunge: I generally prefer to avoid `if` statements without braces – I'd suggest using a `switch` for this sort of thing.

src/php/rest-api/class-snippets-rest-controller.php
```diff
@@ -200,6 +236,72 @@ public function get_items( $request ): WP_REST_Response {
 		$all_snippets = get_snippets( [], $network );
 		$all_snippets = $this->get_network_items( $all_snippets, $network );
 
+		$status = sanitize_key( (string) $request->get_param( 'status' ) );
+
+		if ( in_array( $status, [ 'active', 'inactive' ], true ) ) {
+			$all_snippets = array_filter(
```
sheabunge: Would it be worth combining these `array_filter` calls into a single iteration?

src/php/class-plugin.php
```diff
@@ -121,6 +128,11 @@ public function load_plugin() {
 		$this->evaluate_content = new Evaluate_Content( $this->db );
 		$this->evaluate_functions = new Evaluate_Functions( $this->db );
 
+		// Admin bar integration.
+		require_once $includes_path . '/class-admin-bar.php';
```
sheabunge: Composer autoloader should handle loading this class.

src/php/class-plugin.php
```diff
@@ -121,6 +128,11 @@ public function load_plugin() {
...
+		$this->admin_bar = new Admin_Bar();
+		$this->admin_bar->register_hooks();
```
sheabunge: Might as well combine `register_hooks` into the `Admin_Bar` constructor; there doesn't seem to be really any value to dividing them like this.

sheabunge (approved):

src/php/rest-api/class-snippets-rest-controller.php
imantsk: yes! combined the `array_filter()` calls into a single pass

src/php/class-plugin.php
imantsk: removed

src/php/class-plugin.php
imantsk: You're right! Moved hook registration into `Admin_Bar::__construct()` and removed the extra `register_hooks()` call.

src/js/admin-bar.ts
imantsk: done

src/js/admin-bar.ts
imantsk: fixed

src/js/admin-bar.ts
imantsk: refactored to `switch` + updated `if`s to use braces

PR #336, opened 2026-03-09
Update browsers list db

rami-elementor:


ramiy:
`npx update-browserslist-db@latest`

PR #337, opened 2026-03-10
fix: beta issues

imantsk:
## Summary

This branch fixes the beta issues reported for Code Snippets and includes a set of follow-up admin usability improvements that were discovered while validating those fixes.

### Snippet editor and snippet data
- Restored snippet description rendering in both:
  - the snippet editor view
  - the All Snippets table
- Fixed `Save and Activate` for brand new snippets so activation works reliably on the first save.
- Fixed activation failure handling for snippets with PHP errors:
  - activation does not silently fail
  - the error notice includes the single-line PHP error message and an expandable stack trace
- Added shared notice styling for traced activation errors, including horizontal scrolling for long stack traces.
- Removed the `cs-back` back link from the editor.

### Admin navigation fixes
- Fixed admin bar snippet links to use the correct `id` query parameter.
- Changed the `Edit Snippet` sidebar item so it no longer navigates away from the current page:
  - clicking it focuses the CodeMirror editor

### Export and download fixes
- Fixed bulk export so selected snippets can be exported again as JSON for re-import.
- Kept `Export` and `Download` as separate actions:
  - `Export` produces importable JSON
  - `Download` produces raw snippet files
- Added bulk-only `Download` action for snippet code:
  - single selection downloads as an individual file based on snippet type
  - multiple selections download as a ZIP archive when `ZipArchive` is available
- Adjusted the bulk action UI so:
  - `Apply` is disabled by default and while `Bulk actions` is selected
  - `Apply` enables when a real action is selected

### Flat Files compatibility
- Fixed the Flat Files + LiteSpeed Cache fatal by preventing early bootstrap code from calling unavailable WordPress functions.

### Screen Options
- Added Screen Options support for choosing visible columns on the All Snippets table.
- Added a `Table Options` section for snippets table behavior with a toggle for truncating long row values.
- Restricted Screen Options so:
  - `snippets` view shows `Table Options`, `Columns`, and `Pagination`
  - `cloud-community` shows only `Pagination`

### Table behavior and markup
- Fixed sortable header markup/classes to match WordPress core sorted column behavior.

### Community Cloud pagination
- Applied the Screen Options pagination setting to Community Cloud requests passing `per_page` through the local REST layer and outbound cloud API request.
- Capped cloud `per_page` at `100` even if the user’s screen option is higher (avoids api error messages)

### DIFF
--- a/src/js/components/EditMenu/EditMenu.tsx
+++ b/src/js/components/EditMenu/EditMenu.tsx
@@ -8,7 +8,6 @@ interface EditMenuLinkBinding {
 	originalHref: string | null
 	originalRole: string | null
 	originalTabIndex: string | null
-	originalAriaDisabled: string | null
 	handleClick: (event: MouseEvent) => void
 	handleKeyDown: (event: KeyboardEvent) => void
 }
@@ -58,13 +57,11 @@ const bindEditMenuLink = (menuLink: HTMLAnchorElement, page: string): EditMenuLi
 		originalHref: menuLink.getAttribute('href'),
 		originalRole: menuLink.getAttribute('role'),
 		originalTabIndex: menuLink.getAttribute('tabindex'),
-		originalAriaDisabled: menuLink.getAttribute('aria-disabled'),
 		handleClick,
 		handleKeyDown
 	}
 
 	menuLink.dataset.codeSnippetsDisabled = 'true'
-	menuLink.setAttribute('aria-disabled', 'true')
 	menuLink.setAttribute('role', 'button')
 	menuLink.setAttribute('tabindex', '0')
 	menuLink.classList.add('code-snippets-edit-menu-link')
@@ -80,7 +77,6 @@ const unbindEditMenuLink = ({
 	originalHref,
 	originalRole,
 	originalTabIndex,
-	originalAriaDisabled,
 	handleClick,
 	handleKeyDown
 }: EditMenuLinkBinding) => {
@@ -92,7 +88,6 @@ const unbindEditMenuLink = ({
 	restoreAttribute(menuLink, 'href', originalHref)
 	restoreAttribute(menuLink, 'role', originalRole)
 	restoreAttribute(menuLink, 'tabindex', originalTabIndex)
-	restoreAttribute(menuLink, 'aria-disabled', originalAriaDisabled)
 }
 
 const useEditMenuLinkFocus = () => {
--- a/src/js/components/EditMenu/EditMenu.tsx
+++ b/src/js/components/EditMenu/EditMenu.tsx
@@ -1,5 +1,123 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { SnippetForm } from './SnippetForm'
 
-export const EditMenu = () =>
-	<SnippetForm />
+const EVENT_NAME = 'code_snippets_focus_editor'
+
+interface EditMenuLinkBinding {
+	menuLink: HTMLAnchorElement
+	originalHref: string | null
+	originalRole: string | null
+	originalTabIndex: string | null
+	originalAriaDisabled: string | null
+	handleClick: (event: MouseEvent) => void
+	handleKeyDown: (event: KeyboardEvent) => void
+}
+
+const focusCodeEditor = () => {
+	window.dispatchEvent(new CustomEvent(EVENT_NAME))
+}
+
+const restoreAttribute = (menuLink: HTMLAnchorElement, name: string, value: string | null) => {
+	if (value) {
+		menuLink.setAttribute(name, value)
+		return
+	}
+
+	menuLink.removeAttribute(name)
+}
+
+const getEditPage = (): string | null => {
+	const editUrl = window.CODE_SNIPPETS?.urls.edit
+
+	return editUrl ? new URL(editUrl, window.location.origin).searchParams.get('page') : null
+}
+
+const bindEditMenuLink = (menuLink: HTMLAnchorElement, page: string): EditMenuLinkBinding | undefined => {
+	const menuUrl = new URL(menuLink.href, window.location.origin)
+
+	if (page !== menuUrl.searchParams.get('page')) {
+		return undefined
+	}
+
+	const handleClick = (event: MouseEvent) => {
+		event.preventDefault()
+		focusCodeEditor()
+	}
+
+	const handleKeyDown = (event: KeyboardEvent) => {
+		if ('Enter' !== event.key && ' ' !== event.key) {
+			return
+		}
+
+		event.preventDefault()
+		focusCodeEditor()
+	}
+
+	const binding = {
+		menuLink,
+		originalHref: menuLink.getAttribute('href'),
+		originalRole: menuLink.getAttribute('role'),
+		originalTabIndex: menuLink.getAttribute('tabindex'),
+		originalAriaDisabled: menuLink.getAttribute('aria-disabled'),
+		handleClick,
+		handleKeyDown
+	}
+
+	menuLink.dataset.codeSnippetsDisabled = 'true'
+	menuLink.setAttribute('aria-disabled', 'true')
+	menuLink.setAttribute('role', 'button')
+	menuLink.setAttribute('tabindex', '0')
+	menuLink.classList.add('code-snippets-edit-menu-link')
+	menuLink.removeAttribute('href')
+	menuLink.addEventListener('click', handleClick)
+	menuLink.addEventListener('keydown', handleKeyDown)
+
+	return binding
+}
+
+const unbindEditMenuLink = ({
+	menuLink,
+	originalHref,
+	originalRole,
+	originalTabIndex,
+	originalAriaDisabled,
+	handleClick,
+	handleKeyDown
+}: EditMenuLinkBinding) => {
+	menuLink.removeEventListener('click', handleClick)
+	menuLink.removeEventListener('keydown', handleKeyDown)
+	menuLink.classList.remove('code-snippets-edit-menu-link')
+	delete menuLink.dataset.codeSnippetsDisabled
+
+	restoreAttribute(menuLink, 'href', originalHref)
+	restoreAttribute(menuLink, 'role', originalRole)
+	restoreAttribute(menuLink, 'tabindex', originalTabIndex)
+	restoreAttribute(menuLink, 'aria-disabled', originalAriaDisabled)
+}
+
+const useEditMenuLinkFocus = () => {
+	useEffect(() => {
+		const page = getEditPage()
+
+		if (!page) {
+			return
+		}
+
+		const editMenuLinks = Array.from(document.querySelectorAll<HTMLAnchorElement>('#adminmenu a[href]'))
+			.flatMap(menuLink => {
+				const binding = bindEditMenuLink(menuLink, page)
+
+				return binding ? [binding] : []
+			})
+
+		return () => {
+			editMenuLinks.forEach(unbindEditMenuLink)
+		}
+	}, [])
+}
+
+export const EditMenu = () => {
+	useEditMenuLinkFocus()
+
+	return <SnippetForm />
+}
--- a/src/js/components/EditMenu/EditMenu.tsx
+++ b/src/js/components/EditMenu/EditMenu.tsx
@@ -5,9 +5,9 @@ const EVENT_NAME = 'code_snippets_focus_editor'
 
 interface EditMenuLinkBinding {
 	menuLink: HTMLAnchorElement
-	originalHref: string | null
-	originalRole: string | null
-	originalTabIndex: string | null
+	originalHref: string | undefined
+	originalRole: string | undefined
+	originalTabIndex: string | undefined
 	handleClick: (event: MouseEvent) => void
 	handleKeyDown: (event: KeyboardEvent) => void
 }
@@ -16,8 +16,8 @@ const focusCodeEditor = () => {
 	window.dispatchEvent(new CustomEvent(EVENT_NAME))
 }
 
-const restoreAttribute = (menuLink: HTMLAnchorElement, name: string, value: string | null) => {
-	if (value) {
+const restoreAttribute = (menuLink: HTMLAnchorElement, name: string, value: string | undefined) => {
+	if (undefined !== value) {
 		menuLink.setAttribute(name, value)
 		return
 	}
@@ -25,10 +25,10 @@ const restoreAttribute = (menuLink: HTMLAnchorElement, name: string, value: stri
 	menuLink.removeAttribute(name)
 }
 
-const getEditPage = (): string | null => {
+const getEditPage = (): string | undefined => {
 	const editUrl = window.CODE_SNIPPETS?.urls.edit
 
-	return editUrl ? new URL(editUrl, window.location.origin).searchParams.get('page') : null
+	return editUrl ? new URL(editUrl, window.location.origin).searchParams.get('page') ?? undefined : undefined
 }
 
 const bindEditMenuLink = (menuLink: HTMLAnchorElement, page: string): EditMenuLinkBinding | undefined => {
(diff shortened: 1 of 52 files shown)


src/js/components/common/ListTable/ListTable.tsx
```diff
@@ -94,6 +94,76 @@ const pageItems = <T, >(
 	}
 }
 
+const getVisibleSelected = <T, K extends Key>(visibleItems: T[], getKey: (item: T) => K, selected: Set<K>): Set<K> =>
+	new Set(visibleItems.map(getKey).filter(key => selected.has(key)))
+
+const buildTableNavProps = <K extends Key>({
```
sheabunge: What's the purpose of this function? It seems to just be passing values through directly.

src/js/components/common/ListTable/ListTable.tsx
```diff
@@ -117,28 +187,33 @@ export const ListTable = <T, K extends Key>({
...
-					<TableItems items={visibleItems} {...{ getKey, columns, noItems, setSelected, rowClassName }} />
-				</tbody>
-				<tfoot>
-					<TableHeadings which="foot" {...tableHeadingsProps} />
-				</tfoot>
-			</table>
-			<TableNav which="bottom" {...tableNavProps} />
-		</>
-	)
+	return <ListTableMarkup
```
sheabunge, line 183: What's the benefit in taking this out separately?

src/js/components/common/ListTable/TableHeadings.tsx
```diff
@@ -23,16 +23,22 @@ const SortableHeading = <T, >({
 }: SortableHeadingProps<T>) => {
 	const isCurrent = column.id === sortColumn?.id
 
-	const newSortDirection = isCurrent
+	const nextSortDirection = isCurrent
 		? 'asc' === sortDirection ? 'desc' : 'asc'
 		: column.defaultSortDirection ?? 'asc'
+	const classDirection = isCurrent ? sortDirection : 'asc' === nextSortDirection ? 'desc' : 'asc'
+	const ariaSort = isCurrent ? 'asc' === sortDirection ? 'ascending' : 'descending' : undefined
```
sheabunge: This can be inlined.

src/js/components/common/ListTable/TableItems.tsx
```diff
@@ -34,7 +34,9 @@ interface TableCellProps<T> {
 }
 
 const TableCell = <T, >({ item, column }: TableCellProps<T>) => {
-	const className = `${column.id}-column column-${column.id}`
+	const className = [ `${column.id}-column`, `column-${column.id}`, column.isHidden ? 'hidden' : '' ]
```
sheabunge: Please use the `classnames.classnames` module for this sort of thing.

src/js/components/common/ListTable/TableNav.tsx
```diff
@@ -25,7 +33,9 @@ const BulkActionSelect = <K extends Key>({ which, actions, setSelectedAction }:
 		<select
 			name={`action${'bottom' === which ? '-2' : ''}`}
 			id={`bulk-action-selector-${which}`}
+			value={selectedActionName}
 			onChange={event => {
+				setSelectedActionName(event.target.value)
```
sheabunge, line 38: It's a bit of a shame to lose the typing here – what's the benefit in this change vs what we are losing?

src/js/components/common/ListTable/TableNav.tsx
```diff
@@ -47,9 +57,11 @@ interface BulkActionsProps<K extends Key> extends Required<Pick<TableNavProps<K>
 	disabled?: boolean
 }
 
-const BulkActions = <K extends Key>({ which, actions, applyAction, disabled }: BulkActionsProps<K>) => {
+const BulkActions = function BulkActions<K extends Key>({ which, actions, applyAction, disabled }: BulkActionsProps<K>) {
 	const [selectedAction, setSelectedAction] = useState<ListTableBulkAction<K>>()
+	const [selectedActionName, setSelectedActionName] = useState('-1')
 	const [isPerformingAction, setIsPerformingAction] = useState(false)
+	const isApplyDisabled = !!disabled || isPerformingAction || !selectedAction
```
sheabunge: This can be inlined.

src/js/components/common/ListTable/TableNav.tsx
```diff
@@ -93,29 +106,32 @@ export interface TableNavProps<K extends Key> extends ListTableNavProps<K>, Omit
...
-				/>)}
-
-			{extraTableNav?.(which)}
-			{0 < totalPages && <TablePagination {...{ totalPages, totalItems, which, ...paginationProps }} />}
-
-			<br className="clear" />
-		</div>
-		: null
+}: TableNavProps<K>) {
+	return (
```
sheabunge: Can remove the braces here.

src/js/components/EditMenu/SnippetForm/fields/CodeEditor.tsx
```diff
@@ -14,6 +14,28 @@ interface EditorTextareaProps {
 	textareaRef: RefObject<HTMLTextAreaElement>
 }
 
+const useFocusEditorShortcut = (
+	codeEditorInstance: ReturnType<typeof useSnippetForm>['codeEditorInstance'],
```
sheabunge: Instead of passing these, `useFocusEditorShortcut` can call `useSnippetForm` directly.

src/js/components/EditMenu/EditMenu.tsx
```diff
@@ -1,5 +1,118 @@
...
+const restoreAttribute = (menuLink: HTMLAnchorElement, name: string, value: string | null) => {
+	if (value) {
+		menuLink.setAttribute(name, value)
+		return
+	}
+
+	menuLink.removeAttribute(name)
+}
+
+const getEditPage = (): string | null => {
```
sheabunge: Please use `undefined` instead of `null` where possible – it plays better with TypeScript, and it's best to be consistent.

src/js/hooks/useSubmitSnippet.ts
```diff
@@ -72,6 +73,19 @@ const getSuccessNotice = (
 	}
 }
 
+const getActivationErrorNotice = (snippet: Snippet): ScreenNotice => [
+	'error',
+	__('Snippet could not be activated.', 'code-snippets'),
+	createElement(
```
sheabunge: Why use `createElement` here over JSX?

src/js/utils/files.ts
```diff
@@ -50,3 +54,109 @@ export const downloadSnippetExportFile = (
...
+	network,
+	shared_network,
+	modified,
+	priority,
+	conditionId,
+	lastActive,
+	code_error,
+	code_error_trace
+}: Snippet): SnippetsExport['snippets'][number] => {
+	const exportSnippet: SnippetsExport['snippets'][number] = Object.fromEntries(
```
sheabunge, line 104: This might be cleaner with a literal object constructor instead of a loop/switch (e.g. `parseSnippetObject`)

src/php/Admin/Menus/Manage_Menu.php
```diff
@@ -44,8 +47,10 @@ public function __construct() {
 		}
 
 		add_action( 'admin_menu', array( $this, 'register_upgrade_menu' ), 500 );
+		add_action( 'admin_init', array( $this, 'handle_bulk_download_request' ) );
+		add_action( 'admin_init', array( $this, 'save_truncation_preference' ) );
 		add_filter( 'set-screen-option', array( $this, 'save_screen_option' ), 10, 3 );
-		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_menu_css' ] );
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_menu_css' ) );
```
sheabunge, line 58: Is there a preference for long array syntax?

src/php/Admin/Menus/Manage_Menu.php
```diff
@@ -227,12 +244,24 @@ protected function get_snippets_per_page(): int {
 		$per_page = (int) get_user_option( 'snippets_per_page' );
 
 		if ( empty( $per_page ) || $per_page < 1 ) {
-			$per_page = 999;
+			$per_page = 100;
```
sheabunge: Might be worth taking this out as a constant/introducing a filter hook?

src/php/Admin/Menus/Manage_Menu.php
```diff
@@ -242,6 +271,135 @@ public function render() {
...
+	 * @return string
+	 */
+	public function render_screen_settings( string $screen_settings, \WP_Screen $screen ): string {
+		if ( $this->is_cloud_community_view() ) {
+			return $screen_settings;
+		}
+
+		ob_start();
+		?>
+		<fieldset class="metabox-prefs table-options-prefs">
```
sheabunge, line 362: I wonder if it would be cleaner to construct this within React and reposition the elements instead of relying on so much logic within PHP.

src/php/Admin/Menus/Manage_Menu.php
```diff
@@ -254,4 +412,153 @@ public function render() {
 	public function save_screen_option( $status, string $option, $value ) {
 		return 'snippets_per_page' === $option ? $value : $status;
 	}
+
+	/**
```
sheabunge, line 432: This should be handled through the REST API, IMO.

src/php/Flat_Files/Snippet_Files.php
```diff
@@ -363,7 +363,7 @@ public function sync_active_shared_network_snippets( string $option, $old_value,
 	 * @return void
 	 */
 	public function sync_active_shared_network_snippets_add( $option, $value ): void {
-		if ( 'active_shared_network_snippets' !== $option ) {
```
sheabunge, line 366: huh??

src/php/Flat_Files/Snippet_Files.php
```diff
@@ -400,7 +400,7 @@ private function create_active_shared_network_snippets_file( $value ): void {
 	 * @return string Hashed table name.
 	 */
 	public static function get_hashed_table_name( string $table ): string {
-		return wp_hash( $table );
+		return function_exists( 'wp_hash' ) ? wp_hash( $table ) : md5( $table );
```
sheabunge: wp_hash has [been around since 2.0.3](https://developer.wordpress.org/reference/functions/wp_hash/) – do we really need to assume it may not exist? Or is it a contextual fill-in?

src/php/Model/Model.php
```diff
@@ -105,7 +105,7 @@ function ( $value, $field ) {
 	 * @return string The resolved field name.
 	 */
 	protected static function resolve_field_name( string $field ): string {
-		return self::$field_aliases[ $field ] ?? $field;
+		return static::$field_aliases[ $field ] ?? $field;
```
sheabunge, line 108: Great pick-up, thanks.

src/php/snippet-ops.php
```diff
@@ -755,6 +763,8 @@ function execute_snippet( string $code, int $id = 0, bool $force = false ) {
 		$result = eval( $code );
 	} catch ( ParseError $parse_error ) {
```
sheabunge: We might as well remove the `ParseError` catch if we're going to be catching `Throwable` anyway.

imantsk:
@sheabunge 
ref: https://github.com/codesnippetspro/code-snippets/pull/337#discussion_r2916568675
ref: https://github.com/codesnippetspro/code-snippets/pull/337#discussion_r2916565899

REST downloads are possible (set Content-Disposition: attachment and return raw bytes), but in WP they tend to be more fragile in practice: REST is JSON-first, fetch won’t trigger a native browser download without extra client-side Blob/link work, and large ZIP streaming can run into buffering/memory edge cases. For Screen Options specifically, if we want the controls to live in WP’s native Screen Options flyout (columns/per-page), we’re constrained to core’s PHP hooks (add_screen_option, screen_settings, set-screen-option) and then have React consume/apply the persisted values; moving the UI itself “into React” would mean building and maintaining a separate custom settings UI rather than using native Screen Options.

src/php/Admin/Menus/Manage_Menu.php
imantsk: https://github.com/codesnippetspro/code-snippets/pull/337#issuecomment-4065963083

src/php/Admin/Menus/Manage_Menu.php
imantsk: https://github.com/codesnippetspro/code-snippets/pull/337#issuecomment-4065963083

src/js/components/common/ListTable/ListTable.tsx
imantsk: Removed the redundant pass-through helper and inlined the props construction.
Commit: https://github.com/codesnippetspro/code-snippets/commit/72b6b0db1

src/js/components/common/ListTable/ListTable.tsx
imantsk: removed the unnecessary extraction
Commit: https://github.com/codesnippetspro/code-snippets/commit/72b6b0db1

src/js/components/common/ListTable/TableHeadings.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/b8e442ad5

src/js/components/common/ListTable/TableItems.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/19ed1ea50

src/js/components/common/ListTable/TableNav.tsx
imantsk: fixed 
Commit: https://github.com/codesnippetspro/code-snippets/commit/ecfa584fe

src/js/components/common/ListTable/TableNav.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/ecfa584fe

src/js/components/common/ListTable/TableNav.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/ecfa584fe

src/js/components/EditMenu/SnippetForm/fields/CodeEditor.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/14d0ba5b3

src/js/components/EditMenu/EditMenu.tsx
imantsk: fixed
Commit: https://github.com/codesnippetspro/code-snippets/commit/92c7e01ec

src/js/hooks/useSubmitSnippet.ts
imantsk: Switched to JSX by renaming the hook to .tsx and replacing createElement markup.
Commit: https://github.com/codesnippetspro/code-snippets/commit/3c7afe156

src/js/utils/files.ts
imantsk: Replaced the switch default-check logic with a field-to-check map.
Commit: https://github.com/codesnippetspro/code-snippets/commit/c511d32ee

src/php/Admin/Menus/Manage_Menu.php
imantsk: Added a default-per-page constant + a plugin-specific filter used in both screen option registration and fallback logic.
Commit: https://github.com/codesnippetspro/code-snippets/commit/6f38fe398

src/php/Flat_Files/Snippet_Files.php
imantsk: Rewrote the option handler with a clearer early-return guard style.
Commit: https://github.com/codesnippetspro/code-snippets/commit/0c70958f4

src/php/Flat_Files/Snippet_Files.php
imantsk: it’s because of load order and namespacing, not as much about WP version support.
Plugin like LiteSpeed can trigger this path before pluggable functions are loaded, and in a namespace we must call `\wp_hash`. The guard/fallback is to avoid a fatal during early bootstrap; once WP is fully loaded we’ll use `\wp_hash` as normal.
Commit: https://github.com/codesnippetspro/code-snippets/commit/b9ca72a66

src/php/snippet-ops.php
imantsk: Removed redundant ParseError catches and the unused import.
Commit: https://github.com/codesnippetspro/code-snippets/commit/c11d81954

PR #343, opened 2026-04-10
Fix: accessible title in cloud snippet search results

rami-elementor:
Replace `<a href="#">` with `<button>` to improve semantics. Links are used only for linking to other pages. In this case, clicking the title open a modal. So, it's not a link.

The change also makes keyboard navigation more intuitive. Before you could open only with ENTER, after the change you can use either ENTER or SPACE.

**Before:**

before

**After:**

after

### DIFF
--- a/src/css/manage/_cloud-community.scss
+++ b/src/css/manage/_cloud-community.scss
@@ -81,6 +81,25 @@
 
 	.cloud-snippet {
 		padding: 24px;
+
+		h3 .cloud-snippet-title-button {
+			display: inline;
+			margin: 0;
+			padding: 0;
+			border: none;
+			background: none;
+			font: inherit;
+			color: #2271b1;
+			cursor: pointer;
+			text-align: start;
+			text-decoration: underline;
+		}
+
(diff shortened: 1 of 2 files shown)


louiswol94 (approved):

src/css/manage/_cloud-community.scss
```diff
@@ -81,6 +81,25 @@
...
+			display: inline;
+			margin: 0;
+			padding: 0;
+			border: none;
+			background: none;
+			font: inherit;
+			color: #2271b1;
+			cursor: pointer;
+			text-align: start;
+			text-decoration: underline;
```
imantsk, line 95: what do you say about, no `underline` by default, but only when hover or focus?

rami-elementor: Underlining links is a primary accessibility best practice because it ensures that color is not the only way to identify interactive text.

While the Web Content Accessibility Guidelines (WCAG) do not strictly mandate underlines, they do require that links be visually distinguishable from surrounding text.

I prefer to keep the underline on the link.

sheabunge (approved):

PR #344, opened 2026-04-10
Fix: accessible import snippets tabs

rami-elementor:
Replace `<a href="#">` with `<button>` to improve semantics. Links are used only for linking to other pages. In this case, clicking the element changes the tab. So, it's not a link.

Screenshot 2026-04-10 at 14 03 34

### DIFF
--- a/src/js/components/ImportMenu/ImportMenu.tsx
+++ b/src/js/components/ImportMenu/ImportMenu.tsx
@@ -39,18 +39,17 @@ export const ImportMenu: React.FC = () => {
 			<div className="narrow">
 				<h2 className="nav-tab-wrapper">
 					{TABS.map(tab =>
-						<a
+						<button
 							key={tab}
-							href="#"
+							type="button"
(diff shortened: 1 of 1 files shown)


louiswol94 (approved):

imantsk (approved):

sheabunge:
@rami-elementor The use of placeholder links in this instance was intentional – rather than replacing with buttons, I'd prefer to introduce routing and have the tabs link to actual URLs.

rami-elementor:
@sheabunge For now I prefer to leave it like that.

When you decide to introduce routing and have the tabs links as actual URLs, just refactor the code.

sheabunge:
@rami-elementor my only issue with that is now they are less apparent to someone checking and refactoring – a link to `#` is something that can be searched for and replaced; a `<button>` element fades into the other markup.

I'll just go through and add `TODO` comments for now as a reminder to revisit.

PR #346, opened 2026-04-10
New: migrate from Elementor Pro custom code to Code Snippets

rami-elementor:
Elementor custom code:

image

----

Code Snippets Importer:

image

image

----

Code Snippets:

image

### DIFF
--- a/src/php/REST_API/Import/Plugins/Elementor_Custom_Code_Plugin_Importer.php
+++ b/src/php/REST_API/Import/Plugins/Elementor_Custom_Code_Plugin_Importer.php
@@ -1,0 +1,499 @@
+<?php
+
+namespace Code_Snippets\REST_API\Import\Plugins;
+
+use Code_Snippets\Model\Snippet;
+use WP_Post;
+
+/**
+ * Importer for Elementor Pro Custom Code (and compatible Elementor Custom Code CPT).
+ *
+ * Aligns with Elementor Pro Custom Code metabox: CPT {@see POST_TYPE}, meta keys
+ * {@see META_LOCATION}, {@see META_CODE}, {@see META_PRIORITY_KEYS}. Location values:
+ * `elementor_head`, `elementor_body_start`, `elementor_body_end`.
+ *
+ * @link https://elementor.com/help/custom-code-pro/
+ */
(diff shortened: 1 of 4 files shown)


rami-elementor:
The imported generated with AI, but it's not yet ready for merge. The locations should be tweaked, they are not imported as intended.

louiswol94:
@rami-elementor The scope gap might not be something we can fix within his PR alone - it requires a change to Code Snippets core - adding a `body-open-content` scope. Let's discuss it internally.

louiswol94:
@rami-elementor Coming to think of it, I also had some other locations in the other plugins that were not supported, so I skipped them. But we can definitely add a backlog task to support more locations in Code Snippets.

src/php/REST_API/Import/Plugins/Elementor_Custom_Code_Plugin_Importer.php
```diff
@@ -0,0 +1,499 @@
...
+	 *
+	 * @param array $snippet_data Snippet data object.
+	 * @param bool  $multisite    Whether the snippet is for a multisite network.
+	 *
+	 * @return Snippet|null The created Snippet object or null if unsupported.
+	 */
+	public function create_snippet( array $snippet_data, bool $multisite ): ?Snippet {
+		$kind = $snippet_data['code_kind'] ?? '';
+
+		if ( ! in_array( $kind, [ 'html', 'css', 'js' ], true ) ) {
```
louiswol94, line 135:
```suggestion
		if ( ! in_array( $kind, [ 'html' ], true ) ) {
```
CSS and JS should be in Pro only.

src/php/REST_API/Import/Plugins/Elementor_Custom_Code_Plugin_Importer.php
```diff
@@ -0,0 +1,499 @@
...
+			case 'head':
+				if ( 'html' === $kind ) {
+					return 'head-content';
+				}
+				if ( 'css' === $kind ) {
+					return 'site-css';
+				}
+				if ( 'js' === $kind ) {
+					return 'site-head-js';
+				}
```
louiswol94, line 185:
```suggestion
				if ( 'html' === $kind ) {
					return 'head-content';
				}
```
Same here

PR #355, opened 2026-04-15
Unified Snippets - initial commit

louiswol94:
Plumbing for Unified Snippets feature

### DIFF
--- a/src/php/UnifiedSnippets/Model/Discovered_Snippet.php
+++ b/src/php/UnifiedSnippets/Model/Discovered_Snippet.php
@@ -1,0 +1,195 @@
+<?php
+
+namespace Code_Snippets\UnifiedSnippets\Model;
+
+use Code_Snippets\Model\Model;
+
+/**
+ * Represents a snippet discovered during a site scan.
+ *
+ * Discovered snippets are read-only references to code found elsewhere in the WordPress installation.
+ * They only become managed snippets when explicitly imported.
+ *
+ * @package Code_Snippets
+ *
+ * @property string $hash          Deterministic hash of source + location for deduplication.
+ * @property string $name          Human-readable name (auto-generated or parsed).
+ * @property string $code          The discovered code content.
+ * @property string $type          php, css, js, html, config, or mixed.
+ * @property string $source_type   theme, child-theme, plugin, builder, core, server, customizer, or mu-plugin.
+ * @property string $source_name   Human-readable source (e.g. "Astra", "Elementor", "functions.php").
+ * @property string $source_path   File path, DB reference, or CPT identifier.
+ * @property int    $line_start    Starting line in source file (0 for DB-stored code).
+ * @property int    $line_end      Ending line in source file.
+ * @property bool   $is_active     Whether the code is currently active/running.
+ * @property bool   $is_importable Whether this snippet can be imported into Code Snippets.
+ * @property string $import_notes  Explanation of import constraints.
+ * @property string $risk_level    low, medium, or high.
+ * @property string $checksum      Content hash for change detection between scans.
+ * @property string $scan_date     ISO 8601 timestamp of when this was last scanned.
+ * @property string $scanner_id    Which scanner found this.
+ */
+class Discovered_Snippet extends Model {
+
+	private const VALID_TYPES = [ 'php', 'css', 'js', 'html', 'config', 'mixed' ];
+
+	private const VALID_SOURCE_TYPES = [
+		'theme', 'child-theme', 'plugin', 'builder',
+		'core', 'server', 'customizer', 'mu-plugin',
+	];
+
+	private const VALID_RISK_LEVELS = [ 'low', 'medium', 'high' ];
+
+	/**
+	 * Default values for discovered snippet fields.
+	 *
+	 * @var array<string, mixed>
+	 */
+	protected static array $default_values = [
+		'hash'          => '',
+		'name'          => '',
+		'code'          => '',
+		'type'          => 'php',
+		'source_type'   => '',
+		'source_name'   => '',
+		'source_path'   => '',
+		'line_start'    => 0,
+		'line_end'      => 0,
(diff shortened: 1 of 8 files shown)


src/php/UnifiedSnippets/Model/Discovered_Snippet.php
```diff
@@ -0,0 +1,201 @@
...
+	 * Includes `line_start` and `line_end`. If those change but the extracted code does not
+	 * (e.g. lines shifted above the block), the hash changes while the checksum may not;
+	 * change detection may then report removed + new instead of unchanged. Scanners can avoid
+	 * that by using a stable location key (path + anchor) and omitting volatile line numbers
+	 * from the hash when appropriate.
+	 *
+	 * @return string SHA-256 hash.
+	 */
+	public function generate_hash(): string {
+		$key = implode( '|', [
```
sheabunge, line 149: Should this be idempotent?

src/php/UnifiedSnippets/Model/Discovered_Snippet.php
```diff
@@ -0,0 +1,201 @@
...
+		return $this->hash;
+	}
+
+	/**
+	 * Generate a content checksum for change detection between scans.
+	 *
+	 * @return string MD5 hash of the code content.
+	 */
+	public function generate_checksum(): string {
+		$this->checksum = md5( $this->code );
```
sheabunge, line 168: Should this be idempotent?

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
```diff
@@ -0,0 +1,321 @@
...
+						'description' => 'Optional list of scanner IDs to run. Omit to run all available.',
+						'items'       => [ 'type' => 'string' ],
+					],
+				],
+			]
+		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . static::BASE_ROUTE . '/(?P<scanner_id>[a-z0-9-]+)',
```
sheabunge: `/` is trimmed from both ends of this value, so there's no need to explicitly prepend it.

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
```diff
@@ -0,0 +1,321 @@
...
+	 */
+	public function run_single_scanner( WP_REST_Request $request ) {
+		$scanner_id = $request->get_param( 'scanner_id' );
+		$scanner    = $this->registry->get( $scanner_id );
+
+		if ( ! $scanner ) {
+			return new WP_Error(
+				'scanner_not_found',
+				sprintf( 'Scanner "%s" is not registered.', $scanner_id ),
+				[ 'status' => 404 ]
```
sheabunge, line 215: I wonder if this should be a Bad Request instead of Page Not Found?

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
```diff
@@ -0,0 +1,321 @@
...
+
+		$metadata = $this->store->get_metadata();
+
+		return rest_ensure_response( [
+			'scan_date'   => $metadata['scan_date'],
+			'scanners'    => $metadata['scanners'],
+			'total_count' => $metadata['total_count'],
+			'count'       => count( $snippets ),
+			'snippets'    => array_map(
+				static fn( $s ) => $s->to_array(),
```
sheabunge, line 272: Nice use of `fn`, I forgot it was usable in our min version.

sheabunge (approved):
Looks great! I really appreciate the effort you've put into clean class structures and documentation. 

One note: we might want to hold off on merging to `core-beta` until 4.0 is merged to `core`. It might be worth having a temporary branch to hold this work in the meantime.

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
```diff
@@ -0,0 +1,321 @@
+<?php
+
+namespace Code_Snippets\UnifiedSnippets\REST;
+
```
imantsk: this class should proably go somewehere under `src/php/REST_API/`

louiswol94:
> Looks great! I really appreciate the effort you've put into clean class structures and documentation.
> 
> One note: we might want to hold off on merging to `core-beta` until 4.0 is merged to `core`. It might be worth having a temporary branch to hold this work in the meantime.

@sheabunge Thanks. I will use [feature/unified-snippets](https://github.com/codesnippetspro/code-snippets/tree/feature/unified-snippets) as the base branch and merge all my PR's in there, then we can have one big merge to `core-beta` when we're ready.

src/php/UnifiedSnippets/Model/Discovered_Snippet.php
louiswol94: Yes - same fields produce the same hash/checksum on every call; we can document that or short-circuit if we want to avoid reassigning.

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
louiswol94: I think both are defensible.

404: Treats …/scan/{scanner_id} as identifying a resource; if that scanner isn't registered, the “resource” doesn't exist. This matches common REST style and is easy for clients ("not found").

Note: I already use 400 for scanner_unavailable (registered but not runnable), which could be a nice split.

WDYT?

src/php/UnifiedSnippets/REST/Scan_REST_Controller.php
louiswol94: Thanks. Updated.

louiswol94:
@sheabunge I will merge this for now, so that I can carry on with the feature, but if there is something, please add the comment here and I will address it in the next PR. Thanks!

sheabunge (approved):

PR #356, opened 2026-04-15
Fix: update "what's new" page markup & headings structure

rami-elementor:
Tweak markup and fix headings structure

image


sheabunge (approved):
Looks great! Really nice having someone think about these details.

PR #357, opened 2026-04-15
Fix: remove Enter key handling from ConfirmDialog component

rami-elementor:
The `ConfirmDialog.tsx` captures `onKeyDown` to trigger dialog confirmation.

The issue is that `onKeyDown` captures Enter globally within the modal.

This can conflict with screen reader virtual buffer navigation and activate the confirm action unintentionally.

To avoid unintentional activation, the dialog should capture Enter key only when focus is on the confirm button, or use native button activation behavior.


sheabunge:
We have a similar `onKeyDown` handler for `GenerateCodeModal.tsx` in the Pro version – does that need revising too?

I believe this code was actually inspired by WordPress' experimental [`ConfirmDialog` component](https://developer.wordpress.org/block-editor/reference-guides/components/confirm-dialog/), which is why it might be a bit more janky than other components.

sheabunge (approved):

PR #358, opened 2026-04-17
Fix: improve tooltips accessibility

rami-elementor:
Currently, screen readers never announce tooltip text because it's a purely visual (CSS).

When you hover over the tooltip, it displays it. But it's not reachable for keyboard users.

With this PR, the tooltip can now be reached using your keyboard.

image

In addition, the accessibility tree is more clear:

image

### DIFF
--- a/src/js/components/common/Tooltip.tsx
+++ b/src/js/components/common/Tooltip.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useId } from 'react'
 import classnames from 'classnames'
 import { __ } from '@wordpress/i18n'
 import { trimTrailingChar } from '../../utils/text'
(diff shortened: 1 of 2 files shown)


src/js/components/common/Tooltip.tsx
```diff
@@ -12,21 +12,38 @@ export interface TooltipProps {
...
+	const tooltipId = useId()
+	return (
+		<div className={classnames(
+			'tooltip',
+			{ 'tooltip-block': block, 'tooltip-inline': inline, 'tooltip-start': start, 'tooltip-end': end },
+			className
+		)}>
+			<button
+				type="button"
+				className="tooltip-trigger"
```
sheabunge, line 28: Can we use WordPress' `screen-reader-text` instead of creating our own class name?

rami-elementor: The `.tooltip-trigger` is not used to visually hide the element, it's a style reset for the new `<button>` element.

louiswol94 (approved):

PR #378, opened 2026-05-02
Fix: enhance accessibility in code editor component

rami-elementor:
After the #349 fix, the `jsx-a11y` plugin has a lint error:

image

----

This PR replaces the unsupported `aria-description` attribute with `aria-describedby` on `role=application`.


sheabunge:
Seems pretty straightforward, thanks for the change!

PR #382, opened 2026-05-05
Fix: accessible sort button in a list table

rami-elementor:
Replace `<a href="#">` with a `<button>`.

### DIFF
--- a/src/css/manage/_snippets-table.scss
+++ b/src/css/manage/_snippets-table.scss
@@ -85,9 +85,27 @@
 		list-style: disc;
 	}
 
-	th.sortable a, th.sorted a {
+	th.sortable .list-table-sort-button,
+	th.sorted .list-table-sort-button {
 		display: flex;
 		flex-direction: row;
+		align-items: center;
+		inline-size: 100%;
+		margin: 0;
+		padding: 0;
+		border: none;
+		background: none;
+		font: inherit;
+		color: #2271b1;
+		cursor: pointer;
+		text-align: start;
+	}
+
+	th.sortable .list-table-sort-button:focus-visible,
+	th.sorted .list-table-sort-button:focus-visible {
+		outline: 2px solid #2271b1;
+		outline-offset: 2px;
+		border-radius: 2px;
 	}
 
 	.row-actions {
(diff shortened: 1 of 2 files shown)


sheabunge:
Would this be better kept as a link if we change the sorting variables to sync with query vars, and link it to an actual URL like `&order-by=etc`?

rami-elementor:
Yes but we will have to restructure the entire link logic, currently we have one link for sorting, we will have to create two links for each heading (ASC, DESC). I think it's an over kill.

Besides, the sorting operation is not an external link, it's an action on a table on an existing page. I prefer to keep it as a button.

louiswol94 (commented):
This PR is sound and merge-worthy as an accessibility and semantics fix. And real URLs as a larger follow-up.

src/css/manage/_snippets-table.scss
```diff
@@ -157,7 +175,6 @@
 
 	#wpbody-content & .column-name {
 		white-space: nowrap; /* prevents wrapping of snippet title */
-		vertical-align: top;
```
louiswol94, line 160: `vertical-align: top` removal on `.column-name`: confirm in wp-admin that long titles / wrapped content still align as intended across browsers.

src/css/manage/_snippets-table.scss
```diff
@@ -85,9 +85,27 @@
...
 		display: flex;
 		flex-direction: row;
+		align-items: center;
+		inline-size: 100%;
+		margin: 0;
+		padding: 0;
+		border: none;
+		background: none;
+		font: inherit;
+		color: #2271b1;
```
louiswol94, line 99: Are you sure about the hard coded color `#2271b1`?

louiswol94 (approved):

src/css/manage/_snippets-table.scss
rami-elementor: With `vertical-align: top;`, the "Name" heading alignment is:

Screenshot 2026-05-05 at 22 44 33

----

Without `vertical-align: top;`:

Screenshot 2026-05-05 at 22 44 43

src/css/manage/_snippets-table.scss
rami-elementor: Yes, I'm sure. It's our `$accent` color. We use this color in multiple places.

louiswol94: So why not use the varable `$accent`?

PR #400, opened 2026-06-01
Fix get_item() returning HTTP 500 for a not-found snippet

thisismyurl:
Hey folks,

get_item() in the REST controller returns HTTP 500 when a snippet ID doesn't exist. That should be 404 — 500 means something went wrong on the server, 404 means the resource doesn't exist. The activate_item() and deactivate_item() methods in the same file already return 404 for the same not-found condition, so get_item() was just the odd one out. One digit change.

(full disclosure, I used AI help with the testing and tweaks - Christopher)


sheabunge:
Great catch – thank you for the contribution!

PR #414, opened 2026-07-13
fix: UI review bugs

imantsk:
Integration branch for the admin UI review work. Includes #415 and #422.

### Manage menu structure

- Manage menu responsibilities split into `Manage_Menu_Assets`, `Manage_Menu_Bulk_Download`, `Manage_Menu_Screen_Options` and `Snippet_Type_Counter`.
- Shared `ListTable`, table navigation and pagination components used by both the snippets table and the cloud tables.
- Snippets and community cloud results render in either table or card view; table remains the default.
- Shared SCSS partials for theme tokens, badges, checkboxes, kebab menus, modals, subnav, toolbar and page header.

### Snippet preview modal

- Read-only editor that honours the configured editor theme, with line numbers and a language badge.
- Minimum and maximum dimensions with internal scrolling.
- Falls back to a plain code view when the code editor is unavailable, and loads only the assets that view needs.
- Modal implementation compatible with the minimum supported WordPress version.

### Community cloud

- Snippet descriptions are sanitised in `Cloud_Snippet`, covering every decode path, and rendered as plain text in the card and table views.
- Page numbers returned by the API are converted for display; featured results use the cloud search page size.
- REST collection arguments registered with named keys.
- Row selection added to the cloud table view, and the tab count restored.

### Admin notices

- `Admin\Notice_Filter` removes notices that do not originate from the plugin on plugin screens, with a CSS fallback. Gated by the `code_snippets/admin/filter_foreign_notices` filter, enabled by default.
- Plugin notices carry the `code-snippets-notice` class so version switch, migration and React notices stay visible.

### Accessibility and styling

- Locked badge contrast raised to meet WCAG AA.
- Snippet type badges receive a focus ring in both list and card views.
- Plugin checkbox styling applied consistently, including Screen Options.
- Kebab menu row heights and padding corrected; row action hover colour matched to snippet name links.

### Maintenance

- `stripTags` simplified, with linear handling of comments, malformed tags and block separation.
- Snippet type counts read through `get_snippets()` instead of a direct database query.
- `Model::set_fields()` ignores input that is neither an array nor an object.
- Direct access guards removed from autoloaded menu classes.
- Translator comments used in place of context strings for the cloud author and view count labels.

## Testing

- New PHPUnit coverage for the notice filter, manage menu services, snippet type counter, cloud snippet decoding and preferences REST controller.
- New Playwright specs for the preview modal, notice filtering, badge contrast, cloud featured results and text utilities.
- CI passes: PHPUnit on PHP 7.4 through 8.4, both Playwright projects, and `lint:js`, `lint:styles` and `lint:php`.


imantsk:
Fixed the modal title contract and limited the manage screen to preview-only CodeMirror assets. Verified with lint, PHPUnit, build, and targeted E2E checks.

imantsk:
Guarded the editor bootstrap so the read-only textarea remains a functional fallback when `wp.codeEditor` is unavailable, and added E2E coverage for the disabled syntax-highlighting preference.

imantsk:
Resolved both follow-ups. The fallback E2E setup now explicitly establishes table view before using table selectors, and notice ownership compares normalized plugin-root paths with a directory-separator boundary, including a sibling-prefix regression case. The relevant tests, lint, and build pass.

src/js/utils/Prism.ts
sheabunge, line 1: I believe this is still necessary for the front-end code highlighting?

src/js/utils/text.ts
```diff
@@ -24,7 +24,223 @@ export const truncateWords = (text: string, wordCount: number): string => {
 		: text
 }
 
-export const stripTags = (text: string): string =>
-	text
-		.replace(/<!--[\s\S]*?-->|<\?(?:php)?[\s\S]*?\?>/gi, '')
-		.replace(/<\/?[a-z][a-z0-9]*\b[^>]*>/gi, '')
+// Tags implying a visual break become whitespace so `<p>A</p><p>B</p>`
+// yields 'A B' rather than 'AB'; inline tags are removed without separators.
+const BLOCK_TAG_NAMES = new Set([
```
sheabunge: This seems unnecessarily complex for what it is trying to do. I'm not confident that we can be clear that it is doing its job, especially when it is serving an important security role.

src/php/Admin/Notice_Filter.php
```diff
@@ -0,0 +1,171 @@
...
+namespace Code_Snippets\Admin;
+
+use ReflectionFunction;
+use ReflectionMethod;
+use Throwable;
+use WP_Screen;
+use function Code_Snippets\code_snippets;
+use const Code_Snippets\PLUGIN_FILE;
+
+defined( 'ABSPATH' ) || exit;
```
sheabunge: Unnecessary in a PHP class file.

src/php/Admin/Notice_Filter.php
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	/**
+	 * Print inline styles that hide any residual foreign notices left in the notice region.
+	 *
+	 * @return void
+	 */
+	public function print_fallback_styles() {
+		?>
+		<style>
```
sheabunge, line 53: Why do we need to do all of this reflection processing when we can just hide the notices with CSS?

src/php/Model/Cloud_Snippet.php
```diff
@@ -76,7 +76,9 @@ protected function prepare_field( $value, string $field ) {
 			case 'is_owner':
 				return (bool) $value;
 			case 'description':
-				return ( null === $value ) ? '' : $value;
+				// Remote payloads occasionally carry non-string descriptions;
```
sheabunge: This sort of processing should occur when the remote payload is being decoded, not when the property is being set.

src/css/common/list-table/_layout.scss
```diff
@@ -0,0 +1,264 @@
...
+
+		&:hover {
+			opacity: 1;
+		}
+	}
+}
+
+.active-snippet {
+	td, th {
+		background-color: rgba(#78c8e6, 0.06);
```
ramiy: use modern syntax, `rgb()` instead of `rgba()`.

```suggestion
		background-color: rgb(120 200 230 / 0.06);
```

src/css/common/list-table/_layout.scss
```diff
@@ -0,0 +1,264 @@
...
+	th.sorted .list-table-sort-button {
+		display: flex;
+		flex-direction: row;
+		align-items: center;
+		inline-size: 100%;
+		margin: 0;
+		border: none;
+		background: none;
+		font: inherit;
+		color: #2271b1;
```
ramiy: All `#2271b1` colors can be replaced with sass `theme.$accent`. there are multiple instances that can be changed.

src/css/common/list-table/_pagination.scss
```diff
@@ -0,0 +1,102 @@
...
+	display: flex;
+	align-items: center;
+	gap: 8px;
+	font-size: 14px;
+
+	input[type='checkbox'] {
+		inline-size: 20px;
+		block-size: 20px;
+		margin: 0;
+		border: 1px solid #c3c4c7;
```
ramiy: replace `#c3c4c7` with `theme.$control-border`. Do that in other places too where the corder color is hardcoded.

src/css/manage/_cloud-community-cards.scss
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+// Three equal-height snippet cards per row, matching the manage snippets
+// grid. minmax(0, 1fr) lets columns shrink below their content width so
+// the grid never overflows the page.
+ul.cloud-search-results.code-snippets-cards {
+	grid-template-columns: repeat(3, minmax(0, 1fr));
+
+	@media screen and (width <= 1100px) {
```
ramiy: Remove `screen and ` from the media query.

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,231 @@
...
+						sprintf(_x('%s of ', 'paging', 'code-snippets'), currentPage)}
+					<span className="total-pages">{totalPages}</span>
+				</span>
+			</span>
+		</>
+		: <span className="paging-input">
+			<PagingInput which={which} totalPages={totalPages} {...inputProps} />
+			<span className="tablenav-paging-text">
+				{/* translators: 1: Current page. */
+					_x(' of ', 'paging', 'code-snippets')}
```
ramiy: Not sure how, but the string should be `%1$s of %2$s`.

ramiy: and use `sprintf)`.

src/js/components/ImportMenu/ImportMenu.tsx
```diff
@@ -36,36 +38,39 @@ export const ImportMenu: React.FC = () => {
...
 
-				<WithRestAPIContext>
-					{TABS.map(tab =>
-						<div key={tab} className={classnames('import-snippets-section', { 'active-section': tab === activeTab })}>
-							{TAB_CONTENT[tab]}
-						</div>)}
-				</WithRestAPIContext>
-			</>
+			<div className="snippets-page-header">
+				<h1>{sprintf(
```
ramiy, line 56: `<h1>` ?

src/js/components/ManageMenu/CommunityCloud/CommunityCloud.tsx
```diff
@@ -15,6 +13,11 @@ const TABS = ['snippets', 'bundles'] as const
 
 type CommunityTab = typeof TABS[number]
 
+const TAB_LABELS: Record<CommunityTab, string> = {
+	snippets: __('Snippets', 'code-snippets'),
+	bundles: __('Bundles', 'code-snippets')
+}
+
```
ramiy, line 21: Don't see the advantage of this change.

src/js/components/ManageMenu/CommunityCloud/SearchResult.tsx
```diff
@@ -1,74 +1,106 @@
...
+export const CloudSnippetAuthor: React.FC<CloudSnippetAuthorProps> = ({ snippet }) =>
+	<p className="cloud-snippet-author">
+		<a
+			href={`${window.CODE_SNIPPETS?.urls.cloud}/codevault/${snippet.codevault}`}
+			target="_blank"
+			rel="noopener noreferrer"
+		>
+			{sprintf(
+				/* translators: %s: cloud library author name. */
+				_x('By %s', 'snippet author', 'code-snippets'),
```
ramiy: No need to use both context and translators comment. 

```suggestion
				/* translators: %s: cloud library author name. */
				__('By %s', 'code-snippets'),
```

when using context function `_x()` it creates a separate string, inflating the amount of strings to translate.

src/js/components/ManageMenu/SnippetsTable/SnippetsTableControls.tsx
```diff
@@ -0,0 +1,137 @@
...
+	return 0 < tagsList.size
+		? <div className="alignleft actions">
+			<label htmlFor="snippets-tag-filter" className="screen-reader-text">
+				{__('Filter snippets by tag', 'code-snippets')}
+			</label>
+			<select
+				id="snippets-tag-filter"
+				name="tag"
+				value={currentTag}
+				aria-label={__('Filter snippets by tag', 'code-snippets')}
```
ramiy: `aria-label` is not required, it has a `<label>` pointing to this `<select>` element.

src/php/Core/deactivation-notice.php
```diff
@@ -27,7 +27,9 @@ function code_snippets_deactivation_notice() {
 	$required_php_version = '7.4';
 
 	if ( version_compare( phpversion(), $required_php_version, '<' ) ) {
-		echo '<div class="error fade"><p><strong>';
+		// The code-snippets-notice class exempts this notice from the plugin's
+		// own admin notice filtering on Code Snippets screens.
+		echo '<div class="code-snippets-notice error fade"><p><strong>';
```
ramiy: add `role=region` and `aria-label`.

src/php/Core/deactivation-notice.php
```diff
@@ -46,7 +48,7 @@ function code_snippets_deactivation_notice() {
 	}
 
 	if ( defined( 'CODE_SNIPPETS_FILE' ) ) {
-		echo '<div class="error fade"><p>';
+		echo '<div class="code-snippets-notice error fade"><p>';
```
ramiy: add `role=region` and `aria-label`.

ramiy (changes requested):

src/js/utils/Prism.ts
imantsk: Only the unused admin-side wrapper was removed. The front-end `entries/prism.ts` bundle, its webpack entry, and `Code_Highlighter` are unchanged.

src/js/utils/text.ts
imantsk: Replaced the hand-rolled scanner with a DOMParser-based implementation: parse, drop script/style elements, add block-boundary spacing, take `textContent`, collapse whitespace. The parser handles tag and comment edge cases; the test suite pins the behaviour.

src/js/utils/text.ts
```diff
@@ -24,7 +24,139 @@
...
+		result += text.slice(copied, position) + replacement
+		copied = end
+		position = text.indexOf('<', end)
+	}
+
+	return 0 === copied ? text : result + text.slice(copied)
+}
+
+export const stripTags = (text: string): string => {
+	const withoutComments = text.replace(/<!--[\s\S]*?-->|<\?(?:php)?[\s\S]*?\?>/gi, '')
```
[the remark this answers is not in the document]
imantsk: Resolved by the DOMParser rewrite — no string scanning remains; comment-stripping cases are covered by tests.

src/php/Admin/Notice_Filter.php
imantsk: Removed.

src/php/Admin/Notice_Filter.php
imantsk: Agreed — the reflection unhooking is gone; CSS hiding is now the sole mechanism, and screen detection covers every registered menu hook including Add New.

src/php/Model/Cloud_Snippet.php
imantsk: Moved — descriptions are sanitized where the API payload is decoded, and local model hydration is untouched, with a round-trip regression test.

src/css/common/list-table/_layout.scss
imantsk: Done — `rgb()` syntax in the touched files (stylelint canonicalizes the alpha to `/ 6%`).

src/css/common/list-table/_layout.scss
imantsk: Replaced the hardcoded `#2271b1` instances with `theme.$accent` in the touched stylesheets.

src/css/common/list-table/_pagination.scss
imantsk: Replaced with `theme.$control-border`, along with the other hardcoded border colours in touched files.

src/css/manage/_cloud-community-cards.scss
imantsk: Removed.

src/js/components/common/ListTable/TablePaginationNavigation.tsx
imantsk: The bottom pager now uses `sprintf( _x( '%1$s of %2$s', … ) )`. The top pager keeps split markup because the current page is an editable input (core list-table pattern); its stale translator comment is corrected.

src/js/components/ImportMenu/ImportMenu.tsx
imantsk: Changed to `<h1>`.

src/js/components/ManageMenu/CommunityCloud/CommunityCloud.tsx
imantsk: The map earned its keep since this comment — it is now used in three places (both tab labels and the page title), so removing it would duplicate the strings. Happy to inline it if you still prefer.

src/js/components/ManageMenu/CommunityCloud/SearchResult.tsx
imantsk: Kept the `_x()` context and dropped the redundant translators comment.

src/js/components/ManageMenu/SnippetsTable/SnippetsTableControls.tsx
imantsk: Removed the redundant `aria-label`.

src/php/Core/deactivation-notice.php
imantsk: Added `role="region"` with a translated `aria-label` to both notices.

src/php/Core/deactivation-notice.php
imantsk: Added `role="region"` with a translated `aria-label` here as well.

imantsk:
Addressed the outstanding audit findings: local snippet descriptions now round-trip storage unmodified, with sanitization applied at the cloud payload decode boundary and a regression test covering legacy content; featured results use the cloud-search page size so the `code_snippets/cloud_search/per_page` filter applies consistently; and the cloud table view renders header and per-row selection wired to the existing bulk-download state. Review threads are answered inline. Verified with lint, PHPUnit (single-site and multisite), the webpack build, and targeted Playwright coverage.

imantsk:
Added the omitted block tags to `blockSelector` and covered adjacent-block separation with regression tests (43e13149). Suite passes.

src/js/components/ManageMenu/CommunityCloud/CommunityCloud.tsx
ramiy: OK

src/js/components/ManageMenu/CommunityCloud/SearchResult.tsx
ramiy: we need the translators comment to all the `%s` placeholders. I prefer to remove the context.

ramiy: not sure why the php lint didn't catch this missing translators comment.

imantsk: Applied your suggestion in ba40d96b — `__('By %s')` with the translators comment, and the same treatment for the view count string in `SnippetsTableControls.tsx`, which had a translators comment but kept `_x()` too.

PHPCS never saw either: both are TypeScript, and `WordPress.WP.I18n.MissingTranslatorsComment` only runs over `src/php`. ESLint has no equivalent rule configured here.

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,229 @@
...
+	...buttonProps
+}) =>
+	1 === currentPage
+		? <>
+			<span className="tablenav-pages-navspan button disabled" aria-hidden="true">&laquo;</span>{'\n'}
+			<span className="tablenav-pages-navspan button disabled" aria-hidden="true">&lsaquo;</span>{'\n'}
+		</>
+		: <>
+			<NavigationButton
+				icon={<>&laquo;</>}
```
ramiy, line 65: tested on RTL?

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,229 @@
...
+				newPage={1}
+				className="first-page"
+				/* translators: Hidden accessibility text. */
+				helperText={__('First page', 'code-snippets')}
+				{...buttonProps}
+			/>{'\n'}
+			<NavigationButton
+				icon={<>&lsaquo;</>}
```
ramiy, line 73: tested on RTL?

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+const ForwardNavigationButtons: React.FC<ForwardNavigationButtonsProps> = ({
+	currentPage,
+	totalPages,
+	...buttonProps
+}) =>
+	totalPages === currentPage
+		? <>
+			<span className="tablenav-pages-navspan button disabled" aria-hidden="true">&rsaquo;</span>{'\n'}
+			<span className="tablenav-pages-navspan button disabled" aria-hidden="true">&raquo;</span>
```
ramiy, line 94: same here. RTL.

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,229 @@
...
+		</>
+		: <>
+			<NavigationButton
+				icon={<>&rsaquo;</>}
```
ramiy, line 98: and here.

src/js/components/common/ListTable/TablePaginationNavigation.tsx
```diff
@@ -0,0 +1,229 @@
...
+				newPage={Math.min(totalPages, currentPage + 1)}
+				className="next-page"
+				/* translators: Hidden accessibility text. */
+				helperText={__('Next page', 'code-snippets')}
+				{...buttonProps}
+			/>{'\n'}
+			<NavigationButton
+				icon={<>&raquo;</>}
```
ramiy, line 106: and another one...

ramiy (changes requested):

src/js/components/common/ListTable/TablePaginationNavigation.tsx
imantsk: I ran the admin in RTL (`dir="rtl"` plus WordPress's own RTL stylesheet bundle via `load-styles.php?dir=rtl`) and compared element positions against LTR.

**The arrows are correct.** `«`, `‹`, `›` and `»` are Bidi_Mirrored characters, so the browser mirrors the glyphs itself once the paragraph direction is RTL — the button rendering `»` in the DOM displays as `«`. This is the same reason core's `WP_List_Table::pagination()` hard-codes the identical entities with no `is_rtl()` branch. Verified visually, not just by reasoning: in RTL the last-page button sits leftmost and draws as `«`.

**The layout mirrors too.** I measured start/end offsets for the pagination cluster, view toggles, filters, card header badge, card footer actions, table columns and the preview modal header/footer in both directions — every value matches, because all the new styles use logical properties (`csstools/use-logical: always` is enforced in `stylelint.config.mjs`).

Two real problems in this file that the RTL pass turned up, fixed in 088f1d3f:

- `aria-describedby="table-paging"` pointed at an element that does not exist. Core puts that ID on the `.paging-input` wrapper; we never did, so the reference was dangling. It is now `table-paging-{top,bottom}`, scoped so it stays unique if a screen ever renders two tables.
- `_x(' of ', 'paging')` carried its surrounding spaces inside the translatable string. Now `__('of')` with the spacing in the markup and a mid-sentence translators comment, matching core.

ramiy (approved):

PR #422, opened 2026-07-23
feat: refresh snippet table, preview and notice presentation

imantsk:
### Community Cloud
- show page-level loading and error notices for Community Cloud requests, announced through live regions
- share download state across cards, table rows and the preview so every action reflects a completed download
- add download or edit actions to cloud snippet previews

### Preview modal
- move the type badge into the modal header, and style the modal actions with the plugin control tokens

### Snippet tables
- apply one canonical checkbox style across cards, tables, the toolbar and settings
- line checkbox cells up across the header, footer and both row states, and reserve the active-row accent border so rows no longer shift when toggled
- centre row contents and the activate column, and match column padding between the snippets and cloud tables
- unify row action colours, including hover, and add a focus ring to snippet type badges

### Admin chrome
- hide foreign admin notices on every plugin screen, keeping notices owned by the plugin and by WordPress settings visible
- mark notices rendered by the plugin as its own, so the filter cannot hide them
- match the add new button to the other plugin buttons, and align the upgrade button icon

### Tests
- consolidate the end-to-end coverage added here into fewer page loads, and stub the cloud requests before the first navigation
- cover row alignment, badge focus, cross-mount download state and the notice filtering rules

## Compatibility

- WordPress 6.4 and newer display the preview type badge in the modal header. WordPress 5.5–6.3 ignore the unsupported header action prop; the modal remains functional without the badge.

## Verification

- `npx eslint src tests`
- `npx stylelint "src/css/**/*.scss"`
- `phpcs`
- `npm run build`
- Playwright: full suite across both storage targets, 168 checks

## Note

- The preview modal trash action uses the destructive colours as literals, matching the twenty existing occurrences across the stylesheets. `_theme.scss` has no destructive token yet; adding one and converting every occurrence is tracked separately, so that this change does not leave a token and a literal in use for the same colour.


src/css/common/_modal.scss
```diff
@@ -71,8 +71,9 @@
 	// the header area, clear of the close button.
 	.code-snippets-preview-modal__badge {
 		position: absolute;
-		inset-block-start: 18px;
+		inset-block-start: 35px;
 		inset-inline-end: 56px;
+		transform: translateY(-50%);
```
imantsk: there should be abetter solution imo

imantsk: the absolute positioning and magic offsets are gone. The badge now renders as the first row of the modal content, aligned to the inline end under the title, plain flex flow (f740d765). Placement is covered by the geometry assertions in the preview spec.

imantsk:
Addressed all three findings: the loading and error notices now use live-region semantics (`role="status"` polite / `role="alert"`) so state changes are announced; downloading from the preview refreshes the shared cloud results, so every action for that snippet flips to Edit and repeat downloads are prevented, with e2e coverage; and `releaseRequest` is explicitly typed `() => void` with a zero-argument resolver wrapper. Verified with lint, PHPUnit (single and multisite), the build, and the featured + preview Playwright suites.

imantsk:
Download state is now owned by a shared per-snippet provider rather than each button: card, table, and preview actions read the same pending/downloaded record, so a download in the preview immediately flips every sibling action and repeat downloads are prevented even before the results refresh completes or if it fails. Covered by e2e cases for the delayed-refresh and failed-refresh paths. Verified with lint, PHPUnit (single and multisite), the build, and the featured + preview Playwright suites.

PR #430, opened 2026-08-03
feat: add missing import link to the toolbar upper links

ramiy:
image


ramiy:
> * **Low — src/js/components/common/Toolbar.tsx:46:** The context-free “Import” label can use the wrong grammatical form in translations and is inconsistent with the existing import navigation strings. Use `_x('Import', 'snippets', 'code-snippets')` and import `_x`.

Fixed.

sheabunge:
Good catch!

PR #439, opened 2026-08-04
fix: admin UI and community cloud fixes

imantsk:
Admin UI and community cloud fixes, with matching test coverage.

## Changes

- Snippet preview modal: footer actions gain a re-entrancy-guarded working state, and the modal accepts an `extraActions` render slot for additional footer buttons.
- Community cloud: selection checkboxes and select-all controls are limited to downloadable snippets in both card and table views via a shared `utils/snippets/cloud.ts` helper, with each view owning a single select-all control. The snippets table accepts optional selection state so it can render without bulk actions.
- Cloud search results: card metadata (tags and modified date) renders correctly when tags are present.
- Kebab menu: keyboard navigation no longer intercepts key presses inside form inputs, and the menu list gains vertical padding.
- Snippets table: tag filter select gains an `aria-label`; empty snippet type counts no longer render a stray zero in the subnav.
- Snippet editor: saving a new snippet replaces the history entry instead of pushing a second one.
- Cloud snippet model: string fields coerce scalar values and fall back to an empty string for non-scalar values from remote responses.
- E2e: navigating to the snippets admin page restores the table view when the persisted preference is set to cards; new specs cover settings tab switching and cloud download eligibility.

## Verification

- eslint, phpcs: clean.
- Webpack build: successful.
- PHPUnit: 145 tests, 337 assertions, 0 failures.


src/js/components/common/SnippetPreviewModal.tsx
```diff
@@ -47,19 +56,49 @@ const getPreviewEditorSettings = (type: string): EditorConfiguration => ({
...
-	setIsWorking: (working: boolean) => void
+	working: PreviewWorkingState & { setIsWorking: (working: boolean) => void }
+}
+
+/**
+ * Tracks whether a footer action is in flight. The ref mirrors the state so
+ * `beginWorking` can reject re-entry within the same tick, before React
+ * re-renders with the disabled buttons.
+ */
+const useWorkingState = () => {
```
sheabunge, line 72: This feels incredibly over-engineered – why is this necessary over simply using a single state variable?

src/js/components/EditMenu/SnippetForm/SnippetForm.tsx
```diff
@@ -96,7 +96,7 @@ const EditForm: React.FC<EditFormProps> = ({ children, className }) => {
 							.replace(__('Create New Condition', 'code-snippets'), __('Edit Condition', 'code-snippets'))
 
 						const newUrl = buildUrl(window.CODE_SNIPPETS.urls.edit, { id: response.id })
-						window.history.pushState({}, document.title, newUrl)
+						window.history.replaceState({}, document.title, newUrl)
```
sheabunge, line 99: Why use `replaceState` instead of `pushState`? `pushState` feels more user-friendly given we are changing the page URL itself.

src/js/components/ManageMenu/SnippetsTable/SnippetsTable.tsx
```diff
@@ -51,7 +51,7 @@ const SnippetTypeTab: React.FC<SnippetTypeTabProps> = ({ type, count }) => {
 							<span className="snippet-type-name-short">{__('All', 'code-snippets')}</span>
 						</>}
 				</span>
-				{count && <span className="subnav-count">{count}</span>}
+				{count ? <span className="subnav-count">{count}</span> : null}
```
sheabunge, line 54: Please do not update this – `&&` is cleaner.

src/js/components/ManageMenu/SnippetsTable/SnippetsTableControls.tsx
```diff
@@ -47,9 +47,9 @@ const SnippetStatusCounts = () => {
 						>
 							{`${label} `}
 							<span className="count">{
+								// translators: %d: number of snippets in the current view.
```
sheabunge: Pretty sure these comments need to be on the line beforehand.

src/js/utils/snippets/cloud.ts
sheabunge, line 1: Feels unnecessary to create a whole new utility file for one function.

src/js/components/common/SnippetPreviewModal.tsx
imantsk: A single state variable cannot reject the second click of a double-click: state updates are asynchronous, so both clicks read `isWorking === false` before the re-render disables the buttons, and the action fires twice. The ref mirrors the state synchronously so `beginWorking()` can refuse re-entry within the same tick. Happy to simplify if there is a leaner pattern that covers that case.

src/js/components/EditMenu/SnippetForm/SnippetForm.tsx
imantsk: With `pushState`, pressing back after the first save returns to the `add-snippet` URL, which renders an empty editor for a snippet that now exists — saving again from there creates a duplicate. `replaceState` lets the edit URL take over the history entry, so back leads to the previous page instead of the stale form. Happy to switch it back if keeping the extra history entry is preferred.

sheabunge: Ah fair enough – I suppose that makes sense. Just looking for the best user behaviour here.

PR #445, opened 2026-08-07
Refactor: move all icons to common components

ramiy:
Split grouped icon files into single-icon files. This change consolidate every icon into `src/js/components/common/icons/`, one file per icon (no grouping).

Also, all the icons are now have the same type, and they all accept props to allow custom size and styling.


ramiy:
We can merge this to 3.10, but we don't have to. It can also wait for 4.0.

sheabunge:
Happy to include this in 3.10, but I am curious – what is the benefit of keeping each icon in its own file?

ramiy:
This is the best practice in many React projects, each component in a separate file. Not sure why, I guess it's something to do with tree shaking when bundling (small icons files instead of one giant icons file) and maybe also for code splitting when dynamically loading some routes (importing specific icons instead of large icon files). But I'm not sure. I just know that this is the best practice.

sheabunge:
Interesting! I am unsure if tree shaking applies if we are grouping the icons contextually (only ever importing from ToolbarIcons in the Toolbar file), but happy to enforce this purely for consistency purposes.

PR #446, opened 2026-08-09
Fix: improve checkboxes accessibility for keyboard users

ramiy:


sheabunge:
FWIW, feel free to merge small changes like this that you have personally written or verified without needing PRs.

PR #447, opened 2026-08-09
Fix: update browserslist db version

ramiy:
image


sheabunge:
This caused an issue last time. Have you fully verified that it does not introduce compatibility issues?

ramiy:
@sheabunge I believe last time the compatibility issue caused by the change in `.browserslistrc` file.

PR #453, opened 2026-08-21
chore: migrate from sass variables to css variables

ramiy:
Eventually we will abandon SASS, using native CSS. But for now, the replace SASS variables with CSS variables.

Start review with `_theme.scss` file.

For core/pro compatibility reason, this PR doesn't delete the SASS variables. They will be deleted when we verify Pro no longer uses SASS variables.


sheabunge (approved):
Probably fine to also remove unused variables – they can be removed or retained in Pro when this is merged.

PR #463, opened 2026-08-27
fix: send snippet modification dates with a UTC offset

### LINKED ISSUE #479: Modified column shows UTC time instead of local timezone
## Report

Reported from support: the Modified column in the All Snippets table displays UTC time instead of the site local timezone. On a site at UTC-6, recently saved snippets appear as approximately 6 hours in the future.

Forum reference: https://wordpress.org/support/topic/code-snippets-3-10-0-all-snippets-page-hangs-will-not-load/

## Expected behavior

The Modified column should display times in the site local timezone, matching the WordPress timezone setting.

## Related work

- Pull request: https://github.com/codesnippetspro/code-snippets/pull/463

TallblokeUK:
## The bug

Reported from support against 3.10.0: the **Modified** column shows recently saved snippets in the future — "6 hours from now" on a site at UTC−6, and by whatever the offset is elsewhere. Reproduced exactly.

## Cause

`Snippet::update_modified()` writes `gmdate( 'Y-m-d H:i:s' )`. The value is UTC, but that format carries **no offset**.

Both paths that hand snippets to the browser passed it through verbatim:

- `Snippets_REST_Controller::prepare_item_for_response()` — whose schema already declares `'format' => 'date-time'` and describes the field as "in ISO format", which it wasn't
- the inline `snippetsList` payload in `Manage_Menu_Assets`

The table then calls `humanTimeDiff( snippet.modified )`. An offset-less string is parsed as **local** time, so on any site behind UTC every recent snippet lands in the future by exactly the site's offset.

## The fix

`Snippet::get_modified_iso()` returns the same instant as ISO 8601 with an explicit offset, reusing the existing UTC-aware `get_modified_timestamp()`. Both output paths now send that.

Side benefit: `<time datetime="…">` is now a valid datetime attribute, which it previously wasn't.

**The stored format is unchanged** — nothing about how dates are written, sorted or compared moves. This only affects what goes over the wire.

## Verification

Reproduced the reporter's environment: WordPress 7.0.4, PHP 8.2.33, Code Snippets 3.10.0 (the released zip), 190 snippets, site at UTC−6.

| | before | after |
|---|---|---|
| Modified cell | `6 hours from now` | `21 minutes ago` |
| `datetime` attribute | `2026-08-27 07:35:20` | `2026-08-27T07:35:20+00:00` |

Plus two unit tests covering the ISO output and the null case for a snippet that was never modified. They fail without the change.

PHPUnit: 159 tests, 0 failures. `lint:js` and `lint:styles` clean.

## Note

No changelog entry, consistent with #461 and #462 — didn't want to pick a version ahead of `(Tag): Prepare`.

Fixes #479

### DIFF
--- a/src/php/REST_API/Snippets/Snippets_REST_Controller.php
+++ b/src/php/REST_API/Snippets/Snippets_REST_Controller.php
@@ -826,6 +826,10 @@ final class Snippets_REST_Controller extends REST_Collection_Controller {
 			$response[ $property ] = $item->$property;
 		}
 
+		// The schema declares this as a date-time, so send one: the stored value
+		// is UTC without an offset, which clients read as local time.
+		$response['modified'] = $item->modified_iso;
+
 		return rest_ensure_response( $response );
 	}
 
--- a/src/php/Admin/Menus/Manage/Manage_Menu_Assets.php
+++ b/src/php/Admin/Menus/Manage/Manage_Menu_Assets.php
@@ -111,6 +111,9 @@ class Manage_Menu_Assets {
 				function ( Snippet $snippet ) {
 					$fields = $snippet->get_fields();
 					$fields['code'] = '';
+					// Match the REST response: a UTC value with no offset is read
+					// as local time by the browser.
+					$fields['modified'] = $snippet->modified_iso;
 					return $fields;
 				},
 				get_snippets()
--- a/src/php/Model/Snippet.php
+++ b/src/php/Model/Snippet.php
@@ -464,6 +464,24 @@ class Snippet extends Model {
 	}
 
 	/**
+	 * Retrieve the modification date as an ISO 8601 string, including the UTC
+	 * offset.
+	 *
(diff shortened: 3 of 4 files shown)


src/php/REST_API/Snippets/Snippets_REST_Controller.php
```diff
@@ -826,6 +826,10 @@ public function prepare_item_for_response( $item, $request ) {
 			$response[ $property ] = $item->$property;
 		}
 
+		// The schema declares this as a date-time, so send one: the stored value
+		// is UTC without an offset, which clients read as local time.
+		$response['modified'] = $item->modified_iso;
```
sheabunge, line 831: We control the schema – instead of rewriting variables like this, it would be better to introduce `modified_iso` to the front-end and use that where appropriate.

sheabunge (approved):

ramiy (approved):

PR #465, opened 2026-08-27
fix: apply the Snippets List Order setting again

### LINKED ISSUE #474: Snippets List Order setting has no effect
## Report

Forum topic: https://wordpress.org/support/topic/version-3-10-0-snippets-order-not-working/

Since updating to 3.10.0, the Settings → Snippets List Order setting has no impact on the displayed order. Whatever option is selected, the snippets table does not reflect the chosen sort order.

## Expected behavior

The selected list order option should determine the default sort order of the All Snippets table.

## Related work

- Pull request (core-beta): https://github.com/codesnippetspro/code-snippets/pull/465
- Pull request (3.10.1): https://github.com/codesnippetspro/code-snippets/pull/469

TallblokeUK:
## The bug

Reported from support: **Settings → Snippets List Order** has no effect. Whichever of its five options you pick, the All Snippets table opens in the same order.

Reproduced on two clean installs — WP 7.0.4 / PHP 8.2.33 and WP 7.1 / PHP 8.3.33. `Name (Z-A)` should invert the list; nothing moved on either.

## Cause

The setting is still defined and still rendered, but **nothing reads it**. Its only two references in 3.10.0:

```
Settings_Fields.php:91    'list_order' => 'priority-asc',   ← default value
Settings_Fields.php:191   'list_order' => [ … ],            ← the field in Settings
```

Until 3.10 the list table applied it directly:

```php
// 3.9.2 — class-list-table.php:1311
$order = Settings\get_setting( 'general', 'list_order' );
```

Sorting moved to the column headings during the admin rewrite. The consumer went with the old table; the setting stayed in the UI, reading nothing.

## Approach

The headings are the mechanism now, so this doesn't restore a parallel ordering system. It maps the setting onto the table's **opening sort** — which is what the setting describes itself as: *"Default way to order snippets on the All Snippets admin menu."*

- `list_order` is localized alongside the other manage data
- mapped to a column id and direction (`priority`/`name`/`date`, asc/desc)
- passed to `ListTable` as a new optional `initialSort`

Clicking a heading overrides it for the rest of the visit, exactly as it overrides any other starting order. The `initialSort` prop is optional, so no other `ListTable` consumer changes behaviour.

**On removing it instead:** that was the alternative, and it's defensible given headings supersede it. I didn't, because it's a visible feature removal for anyone who has it configured, and the value is carried in saved settings either way — honouring it costs less than taking it away. Easy to flip if you'd rather retire the setting.

## Verification

Against a reproduction of the reported environment (WP 7.0.4, PHP 8.2.33, 3.10.0, 190 snippets):

| `list_order` | before | after |
|---|---|---|
| `priority-asc` | Repro 001, 002, 003, 004 | Repro 002, 003, 004, 005 |
| `name-desc` | Repro 001, 002, 003, 004 | **Repro 190, 189, 188, 187** |

An e2e test covers it and **fails without the change**. Full suite: 90 Playwright, 157 PHPUnit, 0 failures; `lint:js` and `lint:styles` clean.

`Manage_Menu_Assets_Test` pins the localized key list, so it gains `listOrder`.

## Note

No changelog entry, consistent with #461, #462 and #463.

Fixes #474

### DIFF
--- a/src/js/components/ManageMenu/SnippetsTable/SnippetsListTable.tsx
+++ b/src/js/components/ManageMenu/SnippetsTable/SnippetsListTable.tsx
@@ -12,7 +12,7 @@ import { getTableColumns } from './TableColumns'
 import { useFilteredSnippets } from './WithFilteredSnippetsContext'
 import { INDEX_STATUS, useSnippetsFilters } from './WithSnippetsTableFilters'
 import { BULK_ACTIONS, TRASHED_BULK_ACTIONS, useApplyBulkAction } from './useApplyBulkAction'
-import type { ListTableAction } from '../../common/ListTable'
+import type { ListTableAction, ListTableSortDirection } from '../../common/ListTable'
 import type { SnippetsTableAction } from './useApplyBulkAction'
 import type { Snippet } from '../../../types/Snippet'
 import type { SnippetView } from '../../../types/SnippetView'
@@ -141,6 +141,7 @@ const SnippetsView: React.FC<SnippetsViewProps> = ({
 		: <ListTable
 			items={snippets}
 			getKey={snippet => snippet.id}
+			initialSort={getInitialSort()}
 			className={classnames({ 'truncate-row-values': truncateRowValues })}
 			columns={columns}
 			actions={actions}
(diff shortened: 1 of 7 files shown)


src/js/components/common/ListTable/ListTable.tsx
```diff
@@ -129,20 +129,30 @@ export interface ListTableProps<T, K extends Key, A extends string> extends List
 	items: T[]
 	beforeTable?: ReactNode
 	selectAllControl?: boolean
+
+	/** Column and direction to sort by before the reader touches a heading. */
+	initialSort?: { columnId: string, direction: ListTableSortDirection }
```
sheabunge, line 134: I feel like this is better passed as separate values instead of an object?

src/js/components/ManageMenu/SnippetsTable/SnippetsListTable.tsx
```diff
@@ -155,6 +156,23 @@ const SnippetsView: React.FC<SnippetsViewProps> = ({
 		/>
 }
 
+/**
+ * The "Snippets List Order" setting names a default ordering for the table,
+ * which sorting is now expressed through columns. Map one to the other so the
+ * setting still decides how the list opens; clicking a heading overrides it for
+ * the rest of the visit, as it does for any other starting order.
+ */
+const LIST_ORDER_SORTS: Record<string, { columnId: string, direction: ListTableSortDirection }> = {
```
sheabunge, line 165: No need to encode a table – just partition on `-` and pass the first half as the column ID, the second as the direction.

PR #486, opened 2026-08-29
feat: group settings into task-based tabs

TallblokeUK:
Proposal, opened for discussion rather than to merge as-is. This is visible UI so it needs John's eye before it goes anywhere.

Settings exist in both editions, so per the branching policy this is a shared change: Core first, then sync into Pro.

## Why

Settings have been accumulating in General for a while. On V4 it reaches seventeen fields, and several of them are whole features rather than preferences — minification, file-based execution, performance tracking, security scanning and revisions all sit in the same flat list as *Hide Upgrade Notices*.

The tabs people need when something has gone wrong are also the hidden ones: Version only appears after ticking a box inside Debug, which is a strange place to look when your site has just broken.

## What this does

Groups the settings into tabs by what someone came to the page to do:

| tab | contents |
| --- | --- |
| Editing | snippet fields, code editor, editor preview |
| Running | execution and output |
| Insights | performance, security |
| Library | revisions, cloud |
| Interface | list, admin bar, notices |
| Advanced | access, maintenance |

**A tab with nothing to show does not appear.** One shared map covers both editions: fields belonging to the other edition are simply absent, so free renders four tabs and Pro six, with no separate layout to keep in sync and no hollow tabs.

Long tabs are broken up by group headings. Settings whose description named the control without saying what it was for have been reworded — *"Track Snippet Performance"* is a label, *"Measure how long each snippet takes to run, so you can find the slow one without guessing"* is a reason to switch it on.

## Storage is untouched, deliberately

This is the part worth reviewing closely. Call sites name a section when reading a setting, and the saved option is keyed by section, so physically moving fields between sections would break every read and orphan every existing install.

`Settings_Layout` maps each tab to a list of `[ storage section, field id ]` pairs. `Setting_Field` still receives the **original** section for its `name` attribute and value lookup; only `add_settings_field()` receives the tab. Nothing migrates, and `get_setting( 'general', 'enable_flat_files' )` keeps working untouched.

Verified by hand: unticked a setting from the **Running** tab, saved, and it wrote to `general/`. Stored sections afterwards were unchanged with no new keys.

## Three bugs found on the way

All the same underlying problem, tab identifiers hardcoded in several places:

- **`get_current_section()` defaulted to the literal `'general'`.** Once that section stops existing, opening Settings from the menu renders a page with no settings on it. Now falls back to the first registered tab, which fixes this generally rather than for these six names.
- **`tabs.ts` only refreshed CodeMirror when the section was literally `'editor'`.** CodeMirror measures itself as zero-height while its container is hidden, so the editor preview mounted and rendered blank on any other tab name.
- **`$sections` in `settings.scss`** drives the `[data-active-tab]` show/hide rules and is a hardcoded list, so new tabs rendered into the DOM but stayed invisible.

Tab names now live in the PHP layout, the SCSS list and the JS. Consolidating that is the obvious follow-up.

## Testing

`phpcs`, `stylelint` and `eslint` clean.

Checked on a free install (WP 7.1 / PHP 8.5): no fatal, four tabs shown, Insights and Library correctly absent, group headings render, editor preview mounts at full height, saving round-trips to the original sections.

Also built and checked against V4 on a Pro dev site, where all six tabs populate.

## Next

Once this lands, it syncs into Pro. There is a clickable mockup of the same idea with the reasoning written out, if that is easier to react to than a diff.

### DIFF
--- a/src/php/Settings/Settings_Layout.php
+++ b/src/php/Settings/Settings_Layout.php
@@ -1,0 +1,219 @@
+<?php
+
+namespace Code_Snippets\Settings;
+
+/**
+ * Groups settings fields into tabs for display.
+ *
+ * Storage is deliberately left alone: every value stays in the section it has
+ * always lived in, so `get_setting( 'general', … )` keeps working and nobody's
+ * saved settings are orphaned. This class only decides which tab a field is
+ * drawn under, and in what order.
+ *
+ * The tabs are grouped by what someone came to the page to do, rather than by
+ * which release the setting arrived in. General had grown large enough that
+ * whole features were sitting in the same flat list as small preferences.
+ *
+ * The map below names every field across both editions. Anything not present
+ * in the running install is skipped, and a tab left with nothing in it does not
+ * appear at all, so the free plugin shows fewer tabs than Pro without needing a
+ * separate map.
+ */
+class Settings_Layout {
+
+	/**
+	 * Retrieve every tab, in display order, before availability is considered.
+	 *
+	 * @return array<string, string> Tab identifier keyed to its label.
+	 */
+	public static function get_tabs(): array {
+		$tabs = [
+			'editing'   => __( 'Editing', 'code-snippets' ),
+			'running'   => __( 'Running', 'code-snippets' ),
+			'insights'  => __( 'Insights', 'code-snippets' ),
+			'library'   => __( 'Library', 'code-snippets' ),
+			'interface' => __( 'Interface', 'code-snippets' ),
+			'advanced'  => __( 'Advanced', 'code-snippets' ),
+		];
+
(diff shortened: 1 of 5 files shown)


src/php/Settings/Settings_Layout.php
```diff
@@ -0,0 +1,219 @@
...
+			'running'   => [
+				[ 'general', 'enable_flat_files' ],
+				[ 'general', 'minify_output' ],
+				[ 'general', 'activate_by_default' ],
+			],
+			'insights'  => [
+				[ 'general', 'enable_performance_tracker' ],
+				[ 'general', 'enable_security_scan' ],
+				[ 'general', 'rescan_all_snippets' ],
+			],
```
ramiy, line 79: what is that? it's not visible on the admin screen.

